PR #11, opened 2019-10-31
Update docs to specify required CORS headers

angelcabo:
## Description
The `If-Modified-Since` header is added to every XHR request to disable request caching in IE. (see: https://github.com/o19s/quepid/blob/master/app/assets/javascripts/routes.js#L21) This 
causes CORS preflight request failures due to the default ES `allow-headers` configuration not including the `If-Modified-Since` header.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,7 @@ You will need to configure Elasticsearch to accept requests from the browser usi
 http.cors:
   enabled: true
   allow-origin: /https?:\/\/localhost(:[0-9]+)?/
+  allow-headers: X-Requested-With,Content-Type,Content-Length,if-modified-since
 ```
 
 # Dev Errata


epugh:
Thanks @angelcabo for the PR!   I'm not a CORS expert, so hopefully my understanding is correct.   
If you don't have this `If-Modified-Since` in the `http.cors.allow-headers` list, then you aren't allowed through?  Is this an IE specific issue or any browser?   The reason I ask is because based on your PR, we should also update https://github.com/o19s/quepid/wiki/Troubleshooting-Elasticsearch-and-Quepid#connecting-elasticsearch-to-quepid wiki page.

The part that I am unclear on, is that we have four different ES servers deployed for testing (see https://github.com/o19s/quepid/wiki/Troubleshooting-Elasticsearch-and-Quepid#demo-elasticsearch-indexes-with-the-tmdb-dataset) , and they all use the same `elasticsearch.yml` file:

```
xpack.security.enabled: false
network.host: 0.0.0.0
http.cors.enabled: true
http.cors.allow-origin: "/.*/"
```
Am I right that these four demo ES servers would exhibit the same problem due to not having the `If-Modified-Since` in the list?  

Lastly, does the line care about the case?
```
allow-headers: X-Requested-With,Content-Type,Content-Length,If-Modified-Since
```

angelcabo:
It's strange, I'm having trouble reproducing this now. Sorry, I think a VPN I was on was causing issues that led me to believe it was a CORS related problem based on the flash error message.

All seems to be working fine with the current docs. Thanks for reviewing. I don't think this PR is needed.

epugh:
@angelcabo I'm going to mention in the wiki that VPN's can cause issues...   And otherwise close this PR.   However, big Thank You for submitting it and then testing it afterwords.

PR #24, opened 2019-11-08
Scorers being associated with a team isn't fully supported

epugh:
## Description
Today you can share a scorer wiht a query, a case, or a team.  When you delete a custom scorer, we check if it's part of a query or a case, but we don't with a team.  This adds in a check on that.

Digging in furthur, I saw that you can't remove a scorer from a Team.  Theortically you can delete one, but then you get a foriegn key error with the `teams_scorer` table.   Turns out what we need is a Remove option, similar to a Member.

In Cases, you can archive them, which is the closest to a Delete we get, so we never saw this issue.

## Motivation and Context
I have many custom scorers, and need to clean them out of quepid.com

## How Has This Been Tested?
Tests and manually

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


app/assets/javascripts/components/scorer_listing/scorer_listing.html
```diff
@@ -3,11 +3,14 @@
...
 
     <span class="item-actions">
-      <span ng-if="ctrl.scorer.owned">
-        <share-scorer scorer="ctrl.scorer"></share-scorer>
-        <edit-scorer scorer="ctrl.scorer"></edit-scorer>
-
-        <action-icon title="'Delete'" icon-class="'fa fa-trash'" fn-call="ctrl.deleteScorer()"></action-icon>
+      <share-scorer scorer="ctrl.scorer"></share-scorer>
+      <edit-scorer scorer="ctrl.scorer"></edit-scorer>
+      <span ng-if="ctrl.team">
```
epugh: @worleydl any better way to do an if else here?

epugh:
I am going to merge this soon.

PR #30, opened 2019-11-25
Make Google Analytics optional

CGamesPlay:
## Description
This makes the google analytics and intercom components of Quepid optional. If the environment variables are not configured, the functionality will simply be disabled.

## Motivation and Context
In doing a private installation of Quepid, we have no need for these features, but in the current implementation, these third-party libraries are loaded even when they API keys are not configured.

## How Has This Been Tested?
Running the application and verifying that the JavaScript snippets are removed from the resulting HTML, and no server or client errors are logged.

## Types of changes

## Checklist:

### DIFF
--- a/app.json
+++ b/app.json
@@ -23,10 +23,10 @@
       "required": true
     },
     "INTERCOM_API_KEY": {
-      "required": true
+      "required": false
     },
     "INTERCOM_APP_ID": {
-      "required": true
+      "required": false
     },
     "LANG": {
       "required": true
@@ -41,7 +41,7 @@
       "required": true
     },
     "QUEPID_GA": {
-      "required": true
+      "required": false
     },
     "RACK_ENV": {
       "required": true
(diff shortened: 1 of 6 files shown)


CGamesPlay:
Not sure what the Circle failure is about, but it doesn't look related? Would appreciate any advice.

epugh:
Thanks @CGamesPlay for this!   I'm reviewing it now.....    Periodically it seems like PhantomJS on docker breaks, I *think* we may not be keeping a properly static version of it, and we get breaking changes...

epugh:
Were you able to run `bin/docker r rake test:quepid`, I think the check on the Google Analytics means the tests are failing....   

I'm trying to figure out how to set the `QUEPID_GA` environment variable so the tests run.   Maybe add:

```
ENV['QUEPID_GA'] = "UA-FAKE-GA-CODE-FOR-TESTING"
```
to `test.rb` environment file?

Issue #34, opened 2019-12-02
URL detection treats any occurance of http as a url


pfries:
**Describe the bug**

The new auto-URL feature introduced here https://github.com/o19s/quepid/pull/27 will wrap any string containing the substring "http" anywhere in the string.

**To Reproduce**
1. Add a display field for a document that has http in the value but not a url.
2. Check the display and the entire value is hyperlinked.

**Expected behavior**
Only strings beginning with http should be hyperlinked.

epugh:
And fixed!

PR #35, opened 2019-12-02
Only treat content that begins with "http" as a link

pfries:
Change the regex to detect urls if the first part of the string value is http (ignoring preceding whitespace).

## Motivation and Context

Fixes this issue https://github.com/o19s/quepid/issues/34


epugh:
Thanks!!!

PR #46, opened 2019-12-19
Support auto-complete also for fields used with thumb: prefix

moshebla:
Also make the fields lookup case and whitespace insensitive

I also wanted to make it to select the full field name on enter (currently adds the typed string, and not the full auto-completed field). Might look into that later unless you have an idea.

### DIFF
--- a/app/assets/javascripts/controllers/wizardModal.js
+++ b/app/assets/javascripts/controllers/wizardModal.js
@@ -155,15 +155,28 @@ angular.module('QuepidApp')
       function loadFields (query) {
         var autocompleteList = [];
         var _query = query.trim().toLowerCase();
-        var thumb = false;
-        if (_query.startsWith('thumb:')) {
-            _query = _query.substr(6);
-            thumb = true;
+        var modified = false;
+        var modifierText = '';
+
+        var allowedModifiers = [
+            'media:',
+            'thumb:'
+        ];
+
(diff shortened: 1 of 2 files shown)


epugh:
going to check this out..  I wonder when we should start writing unit tests for this stuff....

epugh:
@moshebla I got side tracked with #48 and didn't get to finish testing this one.   If you get a chance to augment the unit tests, that would be great.   It make take me a bit to grok this code flow ;-)

moshebla:
@epugh sure, can you point me at where the tests for the wizard controller are? couldn't find any

epugh:
I believe this is the right place:

https://github.com/o19s/quepid/blob/master/spec/javascripts/angular/controllers/wizardModal_spec.js

Thanks and happy new year!

epugh:
@moshebla I wanted to let you know I'm attempting to add a test to `wizardModel_spec.js` but am not having much luck...   Let me know if you have more luck! 

```
    it ('lists fields for autocomplete', function() {
      console.log("Running the test");
      expect(true).toBe(true);

      scope.searchFields = ["title","body","image"];

      console.log("Here we go");
      console.log(WizardModalCtrl);

      var autocompleteList = WizardModalCtrl.loadFields("title");

      console.log(autocompleteList);
      expect(autocompleteList.length).toBe(1);

    });
```

epugh:
This looks great, thanks @moshebla and @worleydl

PR #97, opened 2020-03-14
Review defaults for scorers

### LINKED ISSUE #58: Consolidate scorers in schema
**Describe the bug**
Less of a bug and just cleanup to the schema.  Scorers and DefaultScorers overlap each other enough that they can be merged to cleanup some code.  Consolidate the scorers into one `scorers` table with a `default` flag and any other fields that may need be necessary and remove the `default_scorers` table.

**Additional context**
This will require changing the schema then updating the rails and angular bit to make use of one table versus two different ones.

epugh:
Removing some code from scorers that isn't really used, and that is confusing.

## Description
Remove the DEFAULTS concept, since we don't actually ahve a default anymore, cause we are introducing many scorers to Quepid.  So the "default" can't just be assumed.
Also, the frontend custom scorer and default scorer screens use their own set of default scorer code, it doesn't come form the back end.

## Motivation and Context
Fixes #58 at least partly.

## How Has This Been Tested?
unit tests and manually

## Types of changes

- [] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
Need to run a migration.

## Checklist:


epugh:
Assuming this builds succesfully, this is ready for review and commit @worleydl

Issue #185, opened 2020-07-16
Export Queries that don't have judgements


epugh:
**Is your feature request related to a problem? Please describe.**
Today, via the Basic export format, we only export queries that also have documents and ratings:

| query | docid | rating |
| --- | --- | --- |
| Example | 1 | 4 |

If the query `Example` was completely unrated, or if it doesn't match any documents in our corpus, then it won't export.

**Describe the solution you'd like**
I'd like to be able to export all documents in the Basic usecase, as I might be exporting queries to an third party system for rating.   They might not have judgements, or they may not match docs due to terrible search, but I still want to have those queries judged:

| query | docid | rating |
| --- | --- | --- |
| Example | 1 | 4 |
| Unrated Query | 5 | |
| No match Query | | |

**Describe alternatives you've considered**
This could be just the default way that Basic export works...   Or, do we make the current behavior the default, and add a checkbox for "Include unrated queries" that would bring in the _Unrated Query_ and _No match Query_ examples?

**Additional context**
This feature is only useful if you are rating documents outside of Quepid, but creating the queries in Quepid.   However, as part of a ecosystem of products, we want to facilitate that workflow!

epugh:
In poking around, it turns out that if you want to have the docs related to a query, you must SNAPSHOT the case.

epugh:
Fixed in #191.

PR #191, opened 2020-08-25
Export all for basic case

epugh:
## Description
currently when you export basic format, only docs that are rated are included.    However, often you want to export data that hasn't been rated, but does have docs.   The way we do that is to make a snapshot first, and then export that.   The snapshot freezes the list of doc ids.

## Motivation and Context
We want to export query and docs for external system to do judgement rating at scale, outside of Quepid, and then bring it back in.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


epugh:
This fixes #185

PR #194, opened 2020-08-27
Custom query importer as rake task

DmitryKey:
This is a custom query importer implemented as a rake task.
To run it:

1. put the CSV file into db/data directory
2. execute the rake task by providing only the filename (random_100_prod_queries.csv)
    `bin/docker r bin/rake db:custom_query_loader[random_100_prod_queries.csv]`
3. To check, log in under the importer user.
     
Todo:

1. Create a case for each query
2. Implement a mapper from CSV column names into filter queries (for Solr, similar would need to be done for ES)

## Description
This code is drafted based on db/sample_data_seeds.rb.
The idea is to import queries from a given CSV file. Because queries might have their specific parameters, like filter queries, we need to create new case for each query (by design of Quepid -- it allows having only a unique set of parameters for each query in a case).

## Motivation and Context
This change aims to solve the problem of importing queries for manual rating: 

## How Has This Been Tested?
Since it is work in progress, I'm testing by executing the rake task and checking the status of Quepid's cases under the importer user.

## Types of changes

## Checklist:

### DIFF
--- a/db/custom_query_loader.rb
+++ b/db/custom_query_loader.rb
@@ -1,0 +1,76 @@
+######################################
+# Defaults
+######################################
+
+user_defaults = {
+  administrator:    false,
+  agreed:           true,
+  agreed_time:      Time.now,
+  first_login:       false,
+  name:             'No Name',
+  password:         'password',
+  email:            'foo@example.com',
+}
+
+# seed user specifics
+
+def seed_user hash
+  if User.where(email: hash[:email].downcase).exists?
+    User.where(email: hash[:email].downcase).first
+  else
+    User.create hash
+  end
+end
+
+user_specifics = {
+  name:             'Query Importer User',
+  email:            'quepid+qimporter@o19s.com',
+  password:         'quepid+qimporter',
+}
+user_params          = user_defaults.merge(user_specifics)
+tens_of_queries_user = seed_user user_params
+
+require 'csv'
+
+if ENV['CUSTOM_QUERY']
+    path = ENV['PATH']
+    puts "Loading custom queries from path=" + path
+
+    # Ratings
+    puts "Seeding ratings................"
+
+    search_url = "http://quepid-solr.dev.o19s.com:8985/solr/statedecoded/select"
+
+    rails_db = "#{Rails.root}/db/data/"
+    puts rails_db
+    files_list = Dir[rails_db + "*"]
+    puts "File list: " + files_list.join(', ')
+
+    CSV.foreach(rails_db + path, :headers => true) do |row|
+        query = row['keyword']
+        tens_of_queries_case = tens_of_queries_user.cases.first
+        tens_of_queries_case.update case_name: "'" + query[0..185] + "'"
+
+        puts "Creating case: " + tens_of_queries_case.case_name
+
+        opts    = {
+          format:         :hash,
+          force:          true,
+          clear_existing: true,
+        }
+
+        # non-existent doc id: the point is to create a query in this case
+        ratings = [
+                { query_text: query,   doc_id: ' 720784-021190', rating: ' 5' }
+              ]
+
+        puts "Importing query: " + query
+
+        ratings_importer = RatingsImporter.new tens_of_queries_case, ratings, opts
+        ratings_importer.import
+    end
+
+    puts "Done loading queries"
+
+end
+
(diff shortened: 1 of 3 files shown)


epugh:
Thanks for setting up the PR.   I discovered an issue in our CircleCI setup, and it's fixed by this commit: https://github.com/o19s/quepid/commit/8b6987a67e4fe4b0609caca27b9117720e8dd0fc.   If you do a `git cherry-pick 8b6987a67e4fe4b0609caca27b9117720e8dd0fc` then the CI builds should start working again.

epugh:
One thought that I had...   We have had a undocumented feature to add options at the query level...   Accessible at the query from the "Options" button:
query options

This JSON option is accessible in the scorer via the `qOption` hash.   Here is an example of a scorer that @brigaldies wrote:

```
const DEBUG = false;
const SCORE_PASS = 100;
const SCORE_FAIL_QOPTIONS_NOT_FOUND = 1;
const SCORE_FAIL_NUM_FOUND = 2;
let score = SCORE_PASS;

function log(obj) {
    if (DEBUG) {
        console.log(obj);
    }
}

log(docs);

let expectedNumFound;
try {
    expectedNumFound = qOption('count');
    log(`expectedNumFound=${expectedNumFound}`);
    if (expectedNumFound === null) {
        score = SCORE_FAIL_QOPTIONS_NOT_FOUND;
    }
} catch(err) {
    console.log(err);
}

// Test the number of results
if (score === SCORE_PASS && // Don't test if a previous test failed
    numFound() !== expectedNumFound) {
    score = SCORE_FAIL_NUM_FOUND;
}

log(`Final score=${score}`);

setScore(score);
```

What if we made `qOptions` accessible as a curator variable?   Something like:

```
{
  "query": {
    "match": {
      "title": "#$query##",
      "subject": "#$qOptions('subject')##"

    }
  }
}
```

DmitryKey:
thanks for thoughts @epugh!

1. I have merged the upstream changes into my fork and pushed, but not sure that resolved the CI build issue -- and sorry for messing with the commit history in this PR: when I tried to cherry pick (using this guide: https://gist.github.com/ozh/cbce675ba35e0d336cff), it said, that the commit is already on my fork's master. Probably, there is a better way of doing this, but I just merged upstream's master into my branch.

2. For the qOptions: to be honest I love the fact that we can one single query per case: 
  

- This gives visibility into what cases have already been rated
- Using the code in this PR I've loaded 100 queries into the same case -- and that loads for far too long.

If that is possible, how can I programmatically create a new case in rake task? Once this is achieved, we could envision an upload button for a queries file and trigger the rake task, if that makes sense.

epugh:
So looking at https://app.circleci.com/pipelines/github/o19s/quepid/667/workflows/2eccb760-4d72-4503-81a1-f27dda6e5670/jobs/693 it looks like you just have some Ruby linting errors!  Just run `bin/docker r bundle exec rubocop` locally to see them and fix them.  

In terms of creating a new case, whats nice about the Rake task approach is that you are connected to the database, so you can just instantiate the ActiveRecord objects, like `Case.new` and populate them, you don't have to work through the API layer which is nice.   

However, if you are thinking that this is soemthign you want to trigger from an external tool, then I think it might be worth looking back at using the API.   You have the slight awkwardness of authetnicating, however I've used the headless browser appraoch a number of times to automate interacting with Quepid: https://github.com/o19s/agent_q/blob/master/lib/agent_q.rb#L24

DmitryKey:
you rock @epugh !

1. Fixed lint issues -- thanks for the handy command!
2. Implemented per-query case creation. Here is how it looks in my local quepid instance:

Screenshot 2020-08-30 at 12 30 34


What remains to implement from my perspective:

1. Provide some way to map csv headers to loadable fields, like query, case name
2. Add support for per-query Solr/ES parameterization
3. Add feature to select a scorer. Alternatively, users can go and manually change scorers on the created cases
4. Tests
5. Documentation

epugh:
So, for number 3...   If you haven't chosen a scorer, it will use your default scorer (which you can pick).  Just make sure the user that is creating the cases is also the user that picked the default scorer.  Alternatively, just specify the scorer you want...

To play, just do:
> bin/docker r rails c

Then you can do things like:
> Scorer.all.size

or 

> myScorer = Scorer.find_by(name: 'DCG@5')

This is a great way to learn the object model.  I also connect with _Sequel_ on OSX and look at the schema to understand the connections between objects.   Over the past year, I've cleaned up all the table names and columns in the database to follow standard Rails ActiveRecord principles, to make it easier to reason.

PR #238, opened 2020-11-10
control signup enablement via config

tonomonic:
Controls whether or not signing up is permitted.

## Description
Open signup is a potential security risk. This change makes it disable-able (by default it is enabled)
via an environment variable SIGNUP_ENABLED.

## Motivation and Context
To lock down the creation of users to be an administrative function. The administrator
must use the thor script to add users.

## How Has This Been Tested?
1. `bin/docker r bin/rake test:quepid`
1. `bin/docker r bin/rake karma:run`
1. `bin/docker r bundle exec rubocop`
1. SIGNUP_ENABLED not set (in .env)
  navigate to localhost:3000 and verify signup form is present
1. SIGNUP_ENABLED set to false (in .env)
   1. navigate to localhost:3000 and verify signup form is absent
   1. add admin user via `bin/docker r thor user:create -a foo@example.com "Admin" "admin"`, and verify can log in as admin user via UI

Screen Shot 2020-11-10 at 12 11 27 PM

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [ ] I have read the **CONTRIBUTING** document.
   _there is none?_

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,3 +31,6 @@ EMAIL_MARKETING_MODE=false
 
 # Default scorer for new user accounts
 QUEPID_DEFAULT_SCORER=AP@10
+
+# Whether or not signing up via the UI is enabled.
+SIGNUP_ENABLED=true
--- a/app/assets/javascripts/controllers/signup.js
+++ b/app/assets/javascripts/controllers/signup.js
@@ -12,6 +12,7 @@ angular.module('QuepidSecureApp')
       }
 
       $scope.isEmailMarketingMode = configurationSvc.isEmailMarketingMode();
+      $scope.isSignupEnabled = configurationSvc.isSignupEnabled();
 
       $scope.submit = function (agree, emailMarketingAgree, name, email, pass, confirm) {
         $scope.warnAgree    = false;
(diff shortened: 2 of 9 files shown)


epugh:
Thanks @tonomonic for a nicely thought out PR, this should be easily mergable.   One thought I had is should this flag also disable the API end point for registration?  Under the `api/` have a `SignupsController` that could be disabled.   However, if we do that, we would need a way to create users.  Thor works directly with activerecord, by passing the database, so thats fine.  Ideally however, we would add to the `UsersController` a create method that required an authenticated user to create new users...   

If you wanted to add this, happy to wait for that, otherwise, happy to merge this PR as well now.

tonomonic:
@epugh the route to the signups controller is disabled via the change to config/routes.rb. I verified that the signup request sent by the form returns a 404 (before editing the js to hide the signup form).

It would be nice to have a way for admins to add users directly via API or UI, but my preference is to merge this as-is and maybe add that functionality as a future enhancement.

epugh:
Ah, that makes sense.  I've never really done conditional routing like that before.   I'll give it a try later today, and then merge.

tonomonic:
There may be a more Rubyesque way to do it that I'm not aware of. I did
search pretty thoroughly before trying it this way. If you know of another
let me know and I will try it out.

PR #240, opened 2020-11-12
Renaming a Case on the Teams Listing Page doesn't work

epugh:
## Description
Digging in, it turns out that for Cases, we have the `case` object itself have methods like `case.rename(newName)`, however in most other objects we have a Service that manipulates our objects.  So ideally we would have `caseSvc.renameCase(case, newCaseName)`.   

This bug occureed because when we load a `Team` object, we get back a list of `cases`, but we don't add the method `rename(newName)` to those cases, it's just some JSON objects wiht out any of the supporting methods.

## Motivation and Context
Lets pay down some tech debt and start chipping away at the `Case` object, moving it to a pure object without all the methods that interact with the API, and put them in the `caseSvc`.

## Types of changes

- [] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/app/assets/javascripts/components/case_listing/case_listing_controller.js
+++ b/app/assets/javascripts/components/case_listing/case_listing_controller.js
@@ -58,7 +58,6 @@ angular.module('QuepidApp')
         ctrl.clickToEdit.clicked = false;
         if (ctrl.clickToEdit.oldVal !== ctrl.clickToEdit.currVal) {
           ctrl.clickToEdit.oldVal = ctrl.clickToEdit.currVal;
-          //ctrl.thisCase.rename(ctrl.clickToEdit.currVal);
           caseSvc.renameCase(ctrl.thisCase, ctrl.clickToEdit.currVal);
         }
       }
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/components/case_listing/case_listing_controller.js
```diff
@@ -25,11 +27,18 @@ angular.module('QuepidApp')
       ctrl.submit     = submit;
 
       // we may get bound to different cases on moves, reset the state
-      $scope.$watch('thisCase', function() {
-        ctrl.thisCase            = $scope.thisCase;
-        ctrl.clickToEdit.currVal = ctrl.thisCase.caseName.slice(0);
-        ctrl.clickToEdit.clicked = false;
-      });
+      //$scope.$watch('thisCase', function() {
```
epugh, line 30: @worleydl I ended up adding the caseRenamed watch, and it worked in one place and broke elsewhere..  so then I commented it out, and while updates take a ms longer to show, they do and i thought maybe this was safer...

worleydl: This is the only part of the PR that makes me nervous...why are we keeping the commented code around?

epugh: Mostly cause I wanted you to see it!  I will delete now...

PR #241, opened 2020-11-13
WIP: add whitelist for search hosts

tonomonic:
This is a WIP, for early feedback.

Add a whitelist configuration to Quepid that restricts which hosts are
allowed to be used as search backends.

TODO: use the list in the JS code as well, to provide user guidance and
ensure they know early on what the list of valid hosts are.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -34,3 +34,6 @@ QUEPID_DEFAULT_SCORER=AP@10
 
 # Whether or not signing up via the UI is enabled.
 SIGNUP_ENABLED=true
+
+# Whitelist of allowed search hosts
+SEARCH_HOST_WHITELIST=example.com,quepid-elasticsearch.dev.o19s.com
--- a/app/models/try.rb
+++ b/app/models/try.rb
@@ -59,6 +59,9 @@ class Try < ActiveRecord::Base
   # Callbacks
   before_create :set_defaults
 
+  # Validations
+  validates_with ::SearchUrlWhitelistValidator
+
   def args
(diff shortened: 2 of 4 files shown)


epugh:
Looks good!   There are days I regret having a deep rich JS app as well as a server side api ;-).

We should probably have a unit test for the extraction logic, cause I know we at times struggle with that.  

It's good you already know how to extend the `configurationSvc` as the bridge between the server and the js app.

epugh:
when you are ready, remind me, and I'll tweak the settings in Heroku to make this new environment variable set to something so we can test this change.

tonomonic:
I am reconsidering this now. Looks like the server app is mostly used for application data persistence, and that it does not make any calls out to a search engine. I'm not sure if this change actually makes things more secure if all of the search requests come from the JS app. I'll reach back out to the the security folks on my end to determine what they really want as far as ability to lock down the search hosts.

epugh:
Oh wait..  Yeah...   You are right...   <slightly embarrassed grin/>.   Your browser makes all the calls out.

Having said that, if you are trying to control what URLs are on the allowed list when you make the cases, then maybe this would be useful?  

https://quepid.com/images/tour-overview.gif

epugh:
Hey @tonomonic is this still something that makes sense?   Should we close it?

tonomonic:
> Hey @tonomonic is this still something that makes sense? Should we close it?

Nope. Closing it. I will re-open when/if it turns out it is needed.

Issue #225, opened 2020-10-07
AP@10 


epugh:
**Describe the bug**
With AP, we had the query `the big lebowski` where we added `&defType=edismax&mm=100%`, and now get 1 perfect result.  We get a green box, but a score of .20, but should be a score of 1.  

**To Reproduce**
1. Use the TMDB
2. Add `the big lebowski`
3. rate and snapsho.
4. change structure to &defType=edismax&mm=100%
5.  Look at the new score.

**Expected behavior**
Should be a 1 and green.  This assumes that AP@10 with 1 result only is a 1, and not a .10

**Screenshots**

**Additional context**

nathancday:
Coming back to this from @234, but I can no reproduce this behavior. Seems to work as expected, possibly already fixed elsehwere?
<img width="310" alt="Screen Shot 2020-11-09 at 2 22 05 PM" src="https://user-images.githubusercontent.com/808225/98586783-3afd0a80-2297-11eb-9a7b-68c2b5fc6be2.png">

nathancday:
Ignore my comment above, this is a bug that occurs when result sets are less than K/10, I updated the AP@10 score seed code to fix it.

PR #243, opened 2020-11-16
patch up nDCG logic for less than 10 result to fix #234

### LINKED ISSUE #234: nDCG@10 Only Scores once ten Docs are scored
**Describe the bug**
The `nDCG@10` scorer only runs the scorer function once you have rated 10 docs.  If you only have say 3 docs, it never runs.  

**To Reproduce**
1. Go to a case with a query that returns at least 10 docs. 
2. Reset all.
3. Score, and notice that the score doesn't budge off of 0.
4. Once you go from your ninth to your tenth doc, the score then jumps to the right score.

**Expected behavior**
NDCG@10 should score as you rate more docs.

**Additional context**
We did just put out a fix where NDCG@10 had a k of 5!  Now it has a k of 10!

nathancday:
The way we handled scenarios with less than 10 docs was poor so this is a remedy to that.

To address this, we now assume unrated documents get a value defined has `missing_rating` in the code (set to 0) in the default scorerer. So in situations with less than 10 docs rated, we fill/pad out the score array and the ideal score array to a length of 10 with 0's as the fill.

We discussed dropping unrated documents for a condensed list of results to be evaluated by nDCG, for nDCG', but I think this would lead to visual confusions as people use Quepid. Consider the scenarios only position 10 is rated (no other docs are rated) would that be intuitive to see that get a nDCG value of 1? I think the pessimistic default will encourage people to fill in missing ratings.

### DIFF
--- a/db/scorers/ndcg@10.js
+++ b/db/scorers/ndcg@10.js
@@ -1,9 +1,18 @@
 var k = 10 // @Rank
+var missing_rating = 0; // pessimistic assumption
 
-var ideal = topRatings(k)
+var ideal = topRatings(k) // could return less than k if less than k docs have ratings
+var missing_pad = Array(k - ideal.length).fill(missing_rating);
+ideal = ideal.concat(missing_pad);
 
-k = ideal.length < k ? ideal.length : k
-var scores = Array(k).fill(0);
+var scores = Array(k);
+eachDoc(function (doc, i) {
+  if (hasDocRating(i)) {
+    scores[i] = (docRating(i));
+  } else {
+      scores[i] = missing_rating;
+  }
+}, k)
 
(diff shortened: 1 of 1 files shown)


epugh:
When we are ready to merge this, we should make sure to update https://github.com/o19s/quepid/wiki/How-Scoring-Works-in-Quepid then with this change.  I think the pessimistic makes a lot of sense.

epugh:
Also, would we want, if you use ndcg, to maybe surface a message?   Make a very angry hopping frog to encourage you to rate!

epugh:
@nathancday lets look at this on Monday and see if we can merge!

epugh:
This is fixed in PR #253  for issue #225

Issue #244, opened 2020-11-16
Importing ratings from RRE returns 500 error


LiuCao0614:
**Describe the bug**
In my local running Quepid, it returns 500 server error when importing ratings from RRE json file.

**To Reproduce**
1. Spin up a local Quepid
2. Create a new case and connect to a Elastic Search instance
3. In the dashboard, click "Import Ratings" -> "RRE" -> Choose a file -> "Import"
4. "Unable to import ratings from RRE: 500 - Internal Server Error" is returned

**Expected behavior**
RRE json file should be successfully imported

**Screenshots**
![image](https://user-images.githubusercontent.com/22570070/99320652-f508ef00-2820-11eb-9246-f00f2d562dff.png)

epugh:
Ouch!!!!!!   Can you provide a suitability anonymized copy of the RRE formatted file?    This feature has worked for me, but I don't think we've really tested it with a wide variety of RRE sourced files, so it would be great to have yours.   Also, if you look in the README at the section on logs, can you see where it failed?  in dev do `bin/docker r tail -f -n 200 log/development.log` and prod do `bin/docker r tail -f -n 200 log/production.log`

LiuCao0614:
My Quepid is connecting to an internal es instance in my company, so I may not be able to share that. But I did get some logs from Quepid docker:


06:31:29 web.1 | Processing by Api::V1::Import::RatingsController#create as JSON
06:31:29 web.1 | User Exists (0.7ms) SELECT 1 AS one FROM `users` LEFT OUTER JOIN `permissions` ON `permissions`.`user_id` = `users`.`id` WHERE `users`.`id` = 1 LIMIT 1
06:31:29 web.1 | User Load (0.5ms) SELECT `users`.* FROM `users` WHERE `users`.`id` = 1 LIMIT 1
06:31:29 web.1 | Permission Load (0.4ms) SELECT `permissions`.* FROM `permissions` WHERE `permissions`.`user_id` IN (1)
[...]

PR #252, opened 2020-11-21
Harden export import cycle for queries no docs

epugh:
## Description
turns out that the import export cycle is a bit fragile.  Lets make it more robust, and only create ratings in our db if we have both doc and rating pairs.  We also used to not create queries that didn't have docs or rating, and now we proeprly do.  plus we properly export empty queries.

## Motivation and Context
Solves #244 hopefully, though we don't have a smoking gun to prove it!

## How Has This Been Tested?
unit tests and manually

## Types of changes

## Checklist:


epugh:
Okay @LiuCao0614 here is the fix hopefully!   Give it a try...  I will try and release this week a update docker release as well.

Issue #255, opened 2020-11-24
Compare to Snapshot with super deep Chorus case returns 500


epugh:
**Describe the bug**
Can't compare a snapshot.

**To Reproduce**
1. Go to http://app.quepid.com/case/3791/try/NaN?sort=default
2. Click on compare snapshot
3. Pick the baseline
4. Check developer console, we get a 500 from GET | http://app.quepid.com/api/cases/3791/snapshots/1846

**Expected behavior**
Snapshot

**Screenshots**
![Screenshot at Nov 24 14-53-03](https://user-images.githubusercontent.com/22395/100144581-fe1f3f00-2e64-11eb-8d7c-bf18816c691b.png)

**Additional context**

epugh:
Fixed this in commit faf3689db241f626c055c0df9d386493f8db7ec4 which went into the #259 branch accidentally.

Issue #250, opened 2020-11-18
We can Archive a case, but we can't Delete a case, why not?


epugh:
**Is your feature request related to a problem? Please describe.**
I've got old cases that I'll never use again, maybe they were an accident.  I like I can archive and restore old cases, but sometimes I just want to get rid of them.

**Describe the solution you'd like**
Let me Delete a case, not just Archive it.   Prompt me for a confirmation.

**Describe alternatives you've considered**
Don't do anything.

**Additional context**
I've got a LOT of cases now!

epugh:
I really wish I could delete http://app.quepid.com/case/1139?sort=default

PR #262, opened 2020-12-19
Clean up datamodel for case and query

epugh:
## Description
Let cases and queries be deleted.  

## Motivation and Context
I started out looking at #250 about archiving a case being supported, but not deleting.  Digging in, I realized that I don't know *why* we have a `soft_delete` function on Queries.   What is the point of being able to mark a query as deleted, but not actually delete it?   Especially since soft deleting caused a bug in the snapshots, see #255.   As far as I can tell, a soft deleted query is never undeleted..   And never used?   

So branching this out a bit to removing the soft delete function on a query.  If you delete a query you delete the query!

## Types of changes

## Checklist:


epugh:
Looking at the hosted version of Quepid, interestingly we have 23036 deleted queries and only 18461 not deleted queries!

epugh:
BTW, I'm starting to think the soft delete might be useful...   If I have a bunch of queries and ratings, do I want to lose them just because I don't want the query right now on my case?

epugh:
This work was subsumed in first the Rails 5 upgrade, and then in the work on https://github.com/o19s/quepid/pull/288, so this can be closed.

Issue #195, opened 2020-08-28
Rename the "Export Case" and "Import Ratings" to just "Export" and "Import"?


epugh:
<img width="510" alt="Screenshot at Aug 28 12-17-20" src="https://user-images.githubusercontent.com/22395/91590076-c158b000-e928-11ea-8d08-d627818852ee.png">

**Is your feature request related to a problem? Please describe.**
As we've beefed up the Export and Import functions in Quepid to handle roundtripping data better, I notice that the labels that we are using might not be accurate anymore...

"Export Case" lets you export case data, query data, and ratings....   

"Import Ratings" lets you import rating data, but also lets you import just raw set of queries as well.    

**Describe the solution you'd like**
I'm thinking to rename to just *Export* and *Import*...   and move them to next to each other.

**Describe alternatives you've considered**
Thought about "Export Queries" and "Import Queries" or "Export Data" and "Import Data".

epugh:
Argh I just fx this

abhishekChoudhary93:
Can I take this up ?

PR #270, opened 2021-01-21
Renamed "Export Case" and "Import Ratings" to just "Export" and "Import". (#195)

abhishekChoudhary93:
## Description
Renamed "Export Case" and "Import Ratings" to just "Export" and "Import".

## Motivation and Context
#195 

## How Has This Been Tested?
Built changes

## Types of changes

## Checklist:

### DIFF
--- a/app/assets/javascripts/components/export_case/export_case.html
+++ b/app/assets/javascripts/components/export_case/export_case.html
@@ -3,14 +3,14 @@
     <i
       class="fa fa-upload"
       area-hidden="true"
-      title="Export Case"
-      alt="Export Case"
+      title="Export"
+      alt="Export"
     ></i>
   </a>
 </span>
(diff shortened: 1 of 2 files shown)


epugh:
On this ticket, it would be nice if the components `import_ratings` in the source files was renamed to `import` and the `export_case` was renamed to `export`, plus the updates to reference those components as `<import>` instead of `<import-ratings>` in the javascript.    Hence why this is a bit deeper of a ticket!

epugh:
I started making more of these changes, and I finally decided that the title should be short, but hte idea that we are exporting queries and case data and snapshots, well, they are all based on a Case of data, so not remaining the files makes sense.  Plus, having a component with the name `export` felt more confusing.   So going with your changes.

Issue #275, opened 2021-02-18
Adding the possibility to send mails over SMTP


gabauer:
|name           |about                                                            |title                                         |labels|assignees|
|---------------|-----------------------------------------------------------------|----------------------------------------------|------|---------|
|Feature request|Suggest to add the possibility to also use SMTP for sending mails|Adding the possibility to send mails over SMTP|      |         |
---

**Is your feature request related to a problem? Please describe.**

I want to host Quepid for a company for this it is necessary for me that the password reset button is working and can send mails over my own SMTP server. I want to add this functionality and extend the documentation.

**Describe the solution you'd like**

Because I'm new to ruby and rails, I don't know if this is the best solution, but I wanted to add a `smtp_mail_provider.rb` to the initializers folder where if a env var (`MAIL_METHOD`) is present and equals "smtp" the `Rails.application.config.action_mailer.delivery_method` is set to SMTP and the server details are read from the following env vars:

- SMTP_HOST
- SMTP_PORT
- SMTP_USERNAME
- SMTP_PASSWORD
- SMTP_AUTHENTICATION_TYPE
- SMTP_ENABLE_STARTTL

**Describe alternatives you've considered**

An alternative would be to place the code which sets action_mailer.delivery_method to SMTP in the `application.rb` or in an environment file under `environments/`.

- I decided not to use the environment files for this because the code would need to be repeated for test, development, staging, and production.
- I decided not to use the config/application.rb, because I think there should be only defined the standard delivery_method (postmark)

**Additional context**

If it is okay, I would like to take over this ticket.

epugh:
I'd love to see this!   Thanks for sharing your thinking in the issue, this is a nicely written up example!

I think your approach of an initializer makes sense, and setting values based on my reading of https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-configuration.

I expected to find our mailers in `/app/mailers`, and didn't, which reminded me that we use Devise for user auth, and those emails are sent by Devise.   Not sure that matters for waht you are trying!

gabauer:
Thanks for the fast reply! I did the changes regarding this ticket in a local branch, but I can't push the branch to GitHub so that I can open a pull request.

@epugh can you help me with the process of getting permissions to do that?

gabauer:
@epugh please ignore my stupid question. I found out that the normal process is forking and then opening an pull request. Will do this today.

epugh:
There are no stupid questions!   I am super excited about your PR.

PR #276, opened 2021-02-19
Added functionality to send mails over SMTP and start with an operating documentation (#275)

gabauer:
## Description
I added a smtp_mail_provider.rb under `config/initializers`, which checks if the env var `MAIL_METHOD` is set and contains the string "smtp". If this is true, the `Rails.application.config.action_mailer.delivery_method` is overwritten to use smtp for sending mails. Further the following env vars are also added for the smtp settings:

- SMTP_HOST => smtp server to use for sending
- SMTP_PORT =>  smtp port to use for connection
- MAIL_DOMAIN => If you need to specify a HELO domain, you can do set it here
- SMTP_USERNAME => If your mail server requires authentication, set the username in this setting
- SMTP_PASSWORD =>  If your mail server requires authentication, set the password in this setting
- SMTP_AUTHENTICATION_TYPE => If your mail server requires authentication, you need to specify the authentication type here (plain, login, cram_md5)
- SMTP_ENABLE_STARTTLS =>If STARTTLS is enabled in your server set to true

I also added the documentation for this feature. Because there is no really operating documentation, I decided to put one under `docs/operating_documentation.md` and moved the other operating stuff I found in the `README.md` to there.

## Motivation and Context
There are two changes of my POV:

1. Added functionality to send mails over SMTP
This was done because of the issue [here](https://github.com/o19s/quepid/issues/275)

2. Started operating documentation
**Motivation:**
My motivation for this was that an operating documentation did not exists and I planned to extend this documentation in the future (adding how to handle SSL, adding k8s deployment files). In my POV this makes sense because members of search teams mostly don't want to run their own Quepid instance on their machines instead of that they wanted to work as team on one Quepid instance (which is already well supported in Quepid). This central Quepid instance must be operated, which should be made as simple as possible.
**Why this change is required:**
This change is required because of the fact, that in the most companies outside there the department which should run Quepid is not the same as the department which is optimizing the search. In many cases, missing documentation on how to operate Quepid can be a blocker or slow down the process, and we want to see Quepid widespread ;)

## How Has This Been Tested?

- I ran all existing tests.
- I build the Docker image locally and configured SMTP over the new env vars and let send me a password reset mail, which worked.

## Types of changes
- [X] New feature (non-breaking change which adds new functionality)

## Checklist:

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,5 @@
+# Learn more about these settings at https://github.com/o19s/quepid/tree/master/docs/operating_documentation.md
+
 PORT=3000
 
 RACK_ENV=development
@@ -37,3 +39,6 @@ SIGNUP_ENABLED=true
 
 # Whether users are only allowed to use communal (admin controlled) scorers to prevent javascript embedding security issues
 COMMUNAL_SCORERS_ONLY=false
+
+# What provider to use to send emails
+EMAIL_PROVIDER=
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 * Invite your friends to join your team on Quepid!  You can now send an email with an invite code to join Quepid and your specific team.   https://github.com/o19s/quepid/pull/259 by @epugh.
 
+* Add support for sending emails via SMTP, or use Postmark, or don't send emails.  https://github.com/o19s/quepid/pull/276 by @gabaur fixes https://github.com/o19s/quepid/issues/275.
+
 ### Improvements
 
 * Upgrade to Rails 5 and Ruby 2.7.2!   We have been stuck on Rails 4.2 for years, and this unlocks a lot of new capabilities.  https://github.com/o19s/quepid/pull/256 by @epugh with inspiration from @worleydl.
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Below is information related to developing the Quepid open source project, prima
   - [Seed Data](#seed-data)
 - [Data Map](#data-map)
 - [App Structure](#app-structure)
-- [Legal Pages](#legal-pages)
+- [Operating Documentation](#operating-documentation)
 - [Credits](#credits)
 
 <!-- /MarkdownTOC -->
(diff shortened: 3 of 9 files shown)


epugh:
By number of commits, linting issues is the most common ;-)

epugh:
Do you have any sense of what happens if email hasn't been configured at all?  I'm thinking that if we are opening the door to multiple email providers (Postmark, SMTP) then maybe we want to handle the siutation of no email provider set up.   

I'm guessing in that case that we would raise an error...?

epugh:
This is great stuff...  I realized that only the hosted app.quepid.com would be using postmark out of the box, so changing that up to default to NO email provider...   I'm going to test in our staging enviromment if postmark still works.

epugh:
@gabauer I just tested my changes in the staging environment, and they looked good.  Can you give a quick test of the smtp settings, and then I can click merge!

gabauer:
Hi @epugh, I did some tests by hand using Google as SMTP server with the following settings and reset my password, which worked:
- SMTP_HOST=smtp.gmail.com
- SMTP_PORT=587
- MAIL_DOMAIN=localhost
- SMTP_USERNAME=XXXX
- SMTP_PASSWORD=XXXX
- SMTP_AUTHENTICATION_TYPE=plain
- SMTP_ENABLE_STARTTLS=true

I'm pretty sure that's not the kind of test you want. Because I'm really new to ruby I don't know what's the best way to test the functionality, maybe you can help me. If I understood the code correctly, there are already tests regarding the mail content, but not regarding the mail delivery functionality over SMTP which should be done by devise (If configured correctly). Should I add a test based on [rumbster}(https://github.com/sr/rumbster) for example which is testing the pure STMP connecting/sending process?

gabauer:
Hi @epugh, sorry I just realized that I misunderstood your comment. The code with your changes is working for me, so I think you can merge it.

epugh:
Thanks!    If we were doing anything custom to the SMTP process it might be worht a test, but we're assuming that is all standard rails and will work ;-)   Glad to here it owrked.

PR #288, opened 2021-03-12
Actually delete cases

epugh:
## Motivation and Context

Another take on #250.  Archiving is nice, but plenty of times I want to just delete something.. it was a throwaway case...

## Types of changes

## Checklist:


epugh:
So, all of a sudden I am faced with wondering about the orthodoxy that is "All Users Must Have a Case"!   What if we create users and they don't have cases?   Maybe they join a team?   WHat then?

Issue #272, opened 2021-02-02
Expand/Collapse UI Hang


worleydl:
**Describe the bug**
When expanding/collapsing queries on a larger case the UI can become completely locked in Chrome forcing the user to close out the tab.

**To Reproduce**
1. Create a large case
2. Click to expand/collapse queries and use the main paging to navigate around the case until the lock is triggered.

__Note__: Still trying to repro this locally, we did confirm the issue was happening with Liu tho.

**Expected behavior**
User should be able to click around all they want with no lockups

**Screenshots**

**Additional context**

epugh:
I actually want to keep this ticket open, as the workaround isn't how we want to do this!   Hoping we get a real fix!

matteomallus:
Hi @epugh, could you tell me if this issue was resolved or do we still need to set `QUERY_LIST_SORTABLE=true` to prevent the user interface from hanging? I'm also experiencing it on Firefox on Quepid 6.5.5

epugh:
Nothing has been done explicitly to fix this problem, as it was very hard to debug...   Having said that, we are up to 6.14.0, so I'd love to hear if upgrading fixed it..   Lots and lots of good things!

PR #302, opened 2021-03-26
add new env flag to allow query list to be unsortable

LiuCao0614:
## Description

Add a new env flag that if set to false query list are unsortable. 

## Motivation and Context

This mr provides a workaround for issue https://github.com/o19s/quepid/issues/272 before there is a fix. The root cause of this issue is that mousemove event gets triggered first when user clicks the toggle and moves the mouse a little bit. In that case, angular.copy() gets called to save the original order of the query list. However, when the case is big, the cpu usage goes pretty high which causes the UI hanging issue. Suggesting to change the implementation of angular.copy() to async to fix it.

## How Has This Been Tested?

Manually created local docker images with true/false flag and did some smoke tests.

## Types of changes

## Checklist:


epugh:
Thank you for this PR!   Really great piece of debugging of a hard problem.   @binarymax would this solve your issue?

I think my only hesitation with this fix is that it isn't ideal from an end user perspective!  You don't know you will have the problem till you actually run into it...   And then you are annoyed that Quepid broke, until you read enough docs, ask enoguh questions to find this magic setting that "fixes" it ;-)

Here are some alternatives that I see, ordered in my general preference:

1) Can we do some stuff to solve this performance issue?  Solve the root cause?   This would be ideal, though if it requires us to rebuild the front end or really hack Angular, then maybe not viable.


2) Could we identify at runtime when this problem is happening and just pop a message "drag and drop reordering disabled due to number of queries"...?

3) Do we have any sense of how many queries in a case are required to trigger this bug to happen?   We could look to have the sorting capability disable when you reach a certain number of queries in a case.   Honestly, I find the sorting of queries to be very useful in a small case, where I am sorting on the same page.  Once I start paginating, well, we don't support sorting across pages of queries...   We could just gray out the feature with maybe a message?

LiuCao0614:
Yes, like I said this is just a workaround before we find the permanent fix. I think it would be great if we can have this into the next release, so that we can at least set the configuration on our end and avoid the hanging issue.

DmitryKey:
Hey, I've noticed the following link leads to 404: http://quepid-pr-302.herokuapp.com/cases
Screenshot 2021-03-27 at 12 06 32

epugh:
I get in the console instead of /cases, a empty page, and I see this.  I am going to investigate.   Tahnks @DmitryKey  

```
Possibly unhandled rejection: {"data":{"status":500,"error":"Internal Server Error"},"status":500,"config":{"method":"GET","transformRequest":[null],"transformResponse":[null],"jsonpCallbackParam":"callback","url":"/api/users/current","headers":{"Accept":"application/json, text/plain, */*","If-Modified-Since":"0","X-CSRF-TOKEN":"qz7yGi58WjEppM0z9HfrjLtGkSJghs7xcJlnRV0MMdpgiyYiv/lThMmHQ1onwoQj8Db4sR/sspwtzs6EQGQEog==","X-Requested-With":"XMLHttpRequest"}},"statusText":"Internal Server Error","xhrStatus":"complete"}
```

epugh:
Okay, so I think this error is misleading, its because we changed the schema to remove the soft delete function, and since allr eview apps share the staging database, this branch doesn't have that!  hence the error. 

However, i tried on http://quepid-staging.herokuapp.com/case/ which has latest code/latest schema, and it gives an error

epugh:
Btw @LiuCao0614 @worleydl I'd like to cut a new Prod Docker release, and I think this is bascially the only ticket we should wait for at this point.   So if we get this to a mergable state, then I'll cut the new version.

LiuCao0614:
@epugh @worleydl Please let me know if you have any updates on this change. Thanks!

worleydl:
Hey Liu,

We're looking to see if there's a fix that addresses the broken code instead of turning it off.  Still aiming to have something ready this week.

Thanks

LiuCao0614:
Hey Dan,

Would it make sense to have this change in first and roll it back after we find a permanent fix? It would be great if we can have a way to bypass the hanging issue at this point.


-Liu

LiuCao0614:
Hey Eric & Dan,

Could you please take a look this mr and merge this in this week if possible? If would be great if we can update our Quepid running in aws cluster.


Thanks,
Liu

PR #305, opened 2021-03-27
Track create case wizard taken

epugh:
## Description
Working on better logic to trigger the "Create Case Wizard" when you actually create your first case!   Not when you first log in.

## Motivation and Context
We have people joining quepid who are there to Rate!   They aren't goign to create cases, espeically on their very first time...

So, let's redirect those folks to the Teams page, and then when they look at a shared cawe, let them rate.  However, when they click "create case" then pop them into the wizard.

There is still work to do. 

- [x] Rename first_login to completed_create_case_wizard
- [x] Test out the flow, it appears now in this branch first_login never changes!
- [x] Think about if `triggerWizard` should be renamed, and how that logic works.

Already the flow is better!

## How Has This Been Tested?
Manually, and with a test/integration....

## Types of changes

## Checklist:


epugh:
This is ready for testing. on https://quepid-pr-305.herokuapp.com/secure/

PR #350, opened 2021-05-20
Modified nDCG and DCG score calculation as per the defined nDCG Algorithm

lakshmikamsala:
Modified nDCG and DCG score calculation as per the defined nDCG Algorithm defined on 
```
https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Discounted_Cumulative_Gain
```

## Description
I have calculated the nDCG & DCG values manually for the given set of ratings for top 10 records. 
The scores what I got on manual calculations and Quepid nDCG and DCG scores were different.

So, I  tried to modify the scorer algorithm as per the steps given on wiki for DCG calculation.
Modified changes are:
1. Ideal numbers are not in order(desc), So modified code to get it in sequence largest to smallest.
2. Math.Pow function is used while calculating DCG. But, hope it is not required, removed that added just document ratings.

## Motivation and Context
Added above

## How Has This Been Tested?
1. Add few queries.
2. Do the manual calculations as per the formulas defined on wiki.

    **Rank | Rating | log2(rank+1) | Rating/(Log2(Rank+1)) | Ideal Rating | Ideal Rating/log2((Rank+1))**
3. Select the scorer on Quepid and compare DCG and nDCG values.
4. Please find the attached xls for the calculations done on movies test data and the scorers.

Screenshot 2021-05-20 at 6 21 49 PM
[test_movies_detailed.csv](https://github.com/o19s/quepid/files/6516100/test_movies_detailed.csv)
Screenshot 2021-05-20 at 6 14 22 PM
Screenshot 2021-05-20 at 6 14 07 PM

## Types of changes

## Checklist:

### DIFF
--- a/db/scorers/dcg@10.js
+++ b/db/scorers/dcg@10.js
@@ -3,7 +3,7 @@ var score = 0;
 
 eachDoc(function(doc, i) {
     var d = Math.log2(i+2); // i is the JSindex not the DocRank; 0 vs 1
-    var n = Math.pow(2,docRating(i))-1;
+    var n = docRating(i);
     score += d?(n/d):0;
 }, k)
 
--- a/db/scorers/ndcg@10.js
+++ b/db/scorers/ndcg@10.js
@@ -30,9 +30,9 @@ function nDCG(vals, ideal, k) {
   return d ? (n / d) : 0;
 }
 
-let relevance = scores.slice();
-relevance.sort(function(a, b) {
+let ideal_scores = scores.slice();
+ideal_scores.sort(function(a, b) {
   return b - a;
 });
-setScore(nDCG(scores, relevance, k));
+setScore(nDCG(scores, ideal_scores, k));
 
--- a/db/scorers/ndcg@10.js
+++ b/db/scorers/ndcg@10.js
@@ -18,7 +18,7 @@ function DCG(vals, k) {
   var dcg = 0;
   for (var i = 0; i < k; i++) {
     var d = Math.log2(i + 2);
-    var n = Math.pow(2, vals[i]) - 1;
+    var n = vals[i];
     dcg += d ? (n / d) : 0;
   }
   return dcg;
@@ -30,5 +30,9 @@ function nDCG(vals, ideal, k) {
   return d ? (n / d) : 0;
 }
 
-setScore(nDCG(scores, ideal, k));
+let relevance = scores.slice();
+relevance.sort(function(a, b) {
+  return b - a;
+});
+setScore(nDCG(scores, relevance, k));
 


epugh:
I want to review this PR with some other folks, as I want to make sure we are all on the same page of what dcg and ndcg formula means.    There are some various flavours floating around for how to calculate these, and I need to understand how we want to do it, and then make sure we have the documentation to support it!    

We have have a Quepid Qommunity Qoding at 1 PM EST on Friday the 28th, so hoping to go through it then with folks.

nathancday:
There are two forms of DCG (and thus nDCG) defined https://en.wikipedia.org/wiki/Discounted_cumulative_gain. The current code is using the original formulation. The PR code is changing it to the revised formula.

I don't have a preference for which formulation (maybe both?) Quepid supports. I'll defer to community opinion.

db/scorers/ndcg@10.js
```diff
@@ -30,5 +30,9 @@ function nDCG(vals, ideal, k) {
   return d ? (n / d) : 0;
 }
 
-setScore(nDCG(scores, ideal, k));
+let ideal_scores = scores.slice();
```
nathancday, line 33: This will only count docs in `scores`, the top-N windows of results. nDCG requires knowing the global ideal DCG, the best docs available in the entire corpus, not just the top-N. I don't think this code will produce the expected behavior of nDCG.

db/scorers/dcg@10.js
```diff
@@ -3,8 +3,8 @@ var score = 0;
 
 eachDoc(function(doc, i) {
     var d = Math.log2(i+2); // i is the JSindex not the DocRank; 0 vs 1
-    var n = Math.pow(2,docRating(i))-1;
+    var n = docRating(i);
```
nathancday, line 6: this line changes the numerator from the original formulation to an alternative

nathancday (changes requested):
I think the ideal score changes need to be adjusted before merging.

DmitryKey:
As @nathancday says above, Quepid has already implemented the version of DCG, that is referred to as frequently used in the industrial web search engines. Does it equally work for all cases, including non-web search? It depends. Every user should decide for themselves, and what Quepid can offer is different implementations out of the box.

Also: there is always an option to use the scripting feature in the scorers view and implement the variant of the scorer you need!

I did a very quick analysis with different DCG (all non-"industrial" variant, i.e. nominator is the original rating value) with different denominators and two log bases 2 and e:
https://docs.google.com/spreadsheets/d/1VACkpgP7qe9kDidPSiIkDH9LrTCvbm8vsjB6U7xt7OQ/edit?usp=sharing

image

DmitryKey:
hey @lakshmikamsala do you have any feedback on the above conversation?

lakshmikamsala:
> hey @lakshmikamsala do you have any feedback on the above conversation?

Hi @DmitryKey., Sorry., I was sick., I am back to work from today.. 

**As Suggested:** 
"Also: there is always an option to use the scripting feature in the scorers view and implement the variant of the scorer you need!" -> yes, we have option to add our own scoring algorithm.

I have calculated the values as per the  **DCG@P with log2(i+1)** in the above attached XLS. When we look into the Quepid there are no factors defined... So, it might confuse which factors it is following.

epugh:
Would documenting in the scorer all the factors be helpful?  #368 is to enhance the documentation of the scorers.  Maybe eventually we make the "factors" a set of parameters you can tweak on your scorer, though it might just be easier to do that all in code versus building out a whole new scorer customization UI.

DmitryKey:
up. I've drafted the blog post to address this potentially common question -- looking at publishing it as soon as I get a free window.

PR #370, opened 2021-06-03
Fixes SSL termination in Heroku for Issue 271

michaelcizmar:
Small change as described in the issue.  Heroku terminates the SSL connection to Redis so that it will run in Heroku under a Shared Plan.


epugh:
Looking good, want to fix the rubocop lint checks via `bin/docker r bundle exec rubocop --auto-correct-all` ?

Issue #478, opened 2022-03-17
Ancestry is limited to 255 chars and can be exceeded


KennyLindahl:
**Describe the bug**
The `ancestry` column in the `tries` table is limited to **255** chars and can be exceeded.
This bug was discovered when many of our cases lost the ability to click "Rerun queries" in the "Tune relevance" panel.
This results in the `/tries` endpoint returning `500 - internal server error`.

**To Reproduce**
If you currently have just under **255** characters in the `ancestry` field in your `try` and you create another try, then it will crash since the ancestry column is limited to **255** characters.

**Expected behavior**
The `ancestry` column in the `tries` table should either have no limit or should be extended when needed.

**Screenshots**
<img width="760" alt="Screenshot 2022-03-16 at 15 38 52" src="https://user-images.githubusercontent.com/1323068/158759711-389bb26a-be60-41de-bed0-2f7b12b6f37b.png">

<img width="1061" alt="Screenshot 2022-03-17 at 08 32 08" src="https://user-images.githubusercontent.com/1323068/158759734-4d0e0e0b-3e45-4b7e-b67a-36358f7324f2.png">

KennyLindahl:
Probably a duplicate of https://github.com/o19s/quepid/issues/470
But isn't it something we could fix from the Quepid side as suggested in my post above?

epugh:
making some progress @KennyLindahl ;-)    Just went into the db and changed the column defintion to "20" to make testing easy!

Issue #470, opened 2022-02-22
'ancestry' column in 'tries' table has a length constraint


korlowski:
**Describe the bug**
Column `ancestry` in `tries` table has a length constraint (`varchar(255)`), which limits the maximum number of times that a query can be retried while keeping the ancestry. After several dozen retries, clicking on the `Rerun My Searches!` button in the browser does nothing. The console shows a 500 error and the logs show an exception `ActiveRecord::ValueTooLong (Mysql2::Error: Data too long for column 'ancestry' at row 1):`

**To Reproduce**
1. Go to 'Tune Relevance`
2. Keep clicking on `Rerun My Searches!` until the `ancestry` field reaches the limit of 255 characters.
3. Click on `Rerun My Searches!` again
4. Nothing happens and errors are thrown in the console and logs

**Expected behavior**
Queries should be retryable an unlimited number of times.

korlowski:
The workaround is to backtrack manually to an older try (e.g. by changing the URL from `try/65` to `try/1`), but this drops the `ancestry`. Right now this is good enough for us, but the behaviour doesn't seem intended.

epugh:
Have you tried to change the size of the column?   I hadn't ever thought about what happens when the column gets too long, and it sounds like at a minimum we need a nicer error message!   I wonder if we could just bump the size?  (https://www.tutorialspoint.com/What-is-the-maximum-length-of-MySQL-VARCHAR-column)

epugh:
Okay, looking at the website, it appears that the short column is a known thing, and they recommend just making it bigger.  https://github.com/stefankroes/ancestry#internals

Thoughts on resizing the column to something bigger?    

@korlowski, would you be interested in submitting a patch to properly catch the exception and resize the database column?

epugh:
I did a bit of poking, and I didn't realize that the tree structure hits that 255 limit really quickly. Here is what is stored int he database, and see how long my id's are:

```
Try 0: 1048857773
Try 1: 1048857773/1048857774
Try 2: 1048857773/1048857774/1048857775
```

I can't even run 50 tries sequentially...    Now, if you fork etc, then you can.

PR #479, opened 2022-03-28
Ancestry column too short

epugh:
## Description
Modify the table to be bigger.   And, if we generate a new try and it blows up with too long, create a new orphaned try, so you can restart...   At least you can go beyond 200 or so tries!

## Motivation and Context
#470  and #478 have been reported.

## How Has This Been Tested?

unit test and integration test

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/api/v1/tries_controller.rb
+++ b/app/controllers/api/v1/tries_controller.rb
@@ -11,6 +11,8 @@ module Api
         @tries = @case.tries
       end
 
+      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/AbcSize
       def create
         parameters_to_use = try_params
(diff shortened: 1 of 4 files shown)


KennyLindahl:
@epugh Thanks for the fix!

I just tested both the changes:
- The migration changes `ancestry` to `3072` ✅ 
- It resets previous `ancestry` if issue occurs (too long etc) ❌ 
  - It generates the following:
Screenshot 2022-03-29 at 16 18 37

How i tested it was that i changed contents of `ancestry` manually in the DB to length 3071 and then i tried to "Rerun My searches!"
So i suggest that you try the same to reproduce it 😄

epugh:
that is a great suggestion.    My hope was that when we get the too long issue, we would just reset that Try to not having a parent, which would reset the ancestry.

PR #556, opened 2022-08-27
Richer display of scorers

epugh:
## Description
adding in tool tip and description, and more critically, splitting how we calculate the score for "per query" and "per case".  I.e, a score for query, and then your CaseScore is overall....  @david-fisher says we shouldn't just Average with every scorer, that some should be a Sum instead.

## Motivation and Context
Make CaseScore, which I have in the past called your QScore...  your QuepidScore...    More meaningful...

Give newbie users a better sense of which scorer they should reach for!

## How Has This Been Tested?
Still need to work on this.

## Types of changes

- [X ] Breaking change (fix or feature that would cause existing functionality to change)


epugh:
I've added a `sumQueries` method to join the existing `avgQueries` method on the https://github.com/o19s/quepid/blob/richer_display_of_scorers/app/assets/javascripts/controllers/queriesCtrl.js#L211 class...  It's a placeholder and still does averaging!   However, I don't quite understand how that all intersects with the `runScores` method, where the actual logic maybe occurs?  https://github.com/o19s/quepid/blob/richer_display_of_scorers/app/assets/javascripts/controllers/queriesCtrl.js#L134     

I'd love some ideas or fixes @david-fisher @worleydl !

david-fisher:
> I've added a `sumQueries` method to join the existing `avgQueries` method on the https://github.com/o19s/quepid/blob/richer_display_of_scorers/app/assets/javascripts/controllers/queriesCtrl.js#L211 class... It's a placeholder and still does averaging! However, I don't quite understand how that all intersects with the `runScores` method, where the actual logic maybe occurs? https://github.com/o19s/quepid/blob/richer_display_of_scorers/app/assets/javascripts/controllers/queriesCtrl.js#L134
> 
> I'd love some ideas or fixes @david-fisher @worleydl !

I believe the averaging happens in https://github.com/o19s/quepid/blob/richer_display_of_scorers/app/assets/javascripts/services/queriesSvc.js line 1033 when setting the svc.latestScoreInfo inside the scoreAll function (line 991). Angular is not my thing, but that does seem to be the culprit.

epugh:
Bump

PR #557, opened 2022-09-09
Update docker-compose.prod.yml

OkkeKlein:
Changes to make installation guide work with latest version of Quepid

## Motivation and Context
Current installation guide doesn't work for latest version of Quepid as it needs Redis to work.
## How Has This Been Tested?
On a dedicated Ubuntu server with 1Gb memory and 4 Gb swap using a named domain.
## Types of changes

## Checklist:

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -12,15 +12,11 @@ services:
       - MYSQL_ROOT_PASSWORD=password
     volumes:
       - ./volumes/mysql_data:/var/lib/mysql
-
-#  Redis is used to support sending to Google Analytics events.
-#  Enable this if you are running in Docker and have GA setup.
   redis:
     container_name: quepid_prod_redis
     image: redis:6.0.5
     ports:
       - 6379:6379
-
   app:
     container_name: quepid_prod_app
     image: o19s/quepid:latest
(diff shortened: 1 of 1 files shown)


docker-compose.prod.yml
```diff
@@ -15,11 +15,11 @@ services:
 
 #  Redis is used to support sending to Google Analytics events.
```
epugh, line 16: Should we update this comment then?  Maybe "Redis is used for various purposes like Google Analytics events?"...  Or maybe, we should just go ahead and remove these comments altogether.....?

docker-compose.prod.yml
```diff
@@ -52,10 +52,21 @@ services:
       - QUERY_LIST_SORTABLE=true
     command: "foreman s -f Procfile"
     ports:
-      - 80:3000 # Map to port 80 for outside users.
+      - 3000:3000 # Map to port 80 for outside users when you are not using Nginx
```
epugh, line 55: I wonder if exposing port 3000 even makes sense in Prod, when you have nginx?????   Would it make sense to leave the port mapping as `80:3000`, but comment it out, as well as have the comment you added?

epugh (commented):
Thanks for going through this!   I put two thoughts in, and then this will be great...

docker-compose.prod.yml
OkkeKlein: I think you can remove it as you said that more components in future might depend on it.

epugh:
Thanks!

PR #559, opened 2022-09-14
PoC to integrate OpenSearch as the third search engine

mkr:
## Types of changes

## Checklist:


epugh:
image
It should say "Using OS 1" or maybe "Using OpenSearch 1"...    and if you change settings then it bumps it to "Using OpenSearch 2"...

epugh:
image

The explian query dialog shows some additionald debugging info, and it needs a smarter error message!  since this is with OS.

epugh:
@mkr we should think about using the `SearchEngineFilter` in more places to localize which search engine name we are showing in our messages...   We have solr/es hardcoded, and now have added os...   Wonder if we should use this filter instead?

PR #567, opened 2022-10-14
Adding a port to the empty ports list

jzonthemtn:
## Description

Adding the default port to the ports list because if you try to run the `docker-compose.prod.yml` file as-is you will get the error:

```
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.app.ports contains an invalid type, it should be an array
```


epugh (approved):
yeah...   we ran into this annoynace the other day.   

I kind of want to rip out all the nginx complexity and just replace it with CaddyServer in the future.

PR #582, opened 2022-11-18
Rethink how we handle TMDB Demo case settings for various engines

### LINKED ISSUE #580: Case Creation Wizard should have default values that are optimized for TMDB demo, and are simple and sane for your Search Engine
**Describe the bug**
First reported in https://github.com/querqy/chorus/issues/129, where doing a dismax query with `qf=text_all` blows up because the Chorus Solr schema doesn't have a `text_all`.    

**Expected behavior**
I dug into how Quepid works, and we have a bit of a mishmash of "Are you doing the TMDB example for Solr/OS/ES, if so let me give you a better starting experience" versus a "I know nothing about your search engine, so lets go with most basic settings".   

We should rework how Quepid case wizard works so that we provide a great optimized experience when you pick one of our TMDB demo servers, but if not, then we don't make a bunch of assumptions that might or might not work.

**Screenshots**

**Additional context**
Once this is in, please update Chorus ;-)

epugh:
## Motivation and Context
Fixes #580 

## How Has This Been Tested?
manually mostly

## Types of changes

## Checklist:


epugh:
I still need a simplified ES and OS query to put in here, and then can merge.

PR #594, opened 2022-12-13
Unshare case

### LINKED ISSUE #564: Add "Unshare case" feature
**Is your feature request related to a problem? Please describe.**
When drafting query cases, I can start sharing them with a team, but later change my mind.
Currently Quepid only allows one-way sharing, but no unsharing functionality is available.

**Describe the solution you'd like**
Flip "Share case" to "Unsare case" if it is shared.

**Describe alternatives you've considered**
First cloning, then deleting the shared case could be one workaround.

epugh:
## Description
Fix #564 

## Motivation and Context
you can share, but what if you goofed?   Let's make it possible to unshare!

## How Has This Been Tested?
manually

### DIFF
--- a/app/assets/javascripts/components/share_case/_modal.html
+++ b/app/assets/javascripts/components/share_case/_modal.html
@@ -40,7 +40,7 @@
     <h4>Already shared with:</h4>
 
     <ul class="list-group">
-      <li class="list-group-item list-group-item-success" ng-repeat="team in ctrl.share.sharedTeams">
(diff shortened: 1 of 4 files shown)


epugh:
@DmitryKey before I merge this, want to give it a test?  Heroku should provide a link to it ;-)

DmitryKey:
Hey @epugh, Happy New Year!

Very cool, thanks for crafting it!
I've prepared a test setup on prod and am ready to check it out when it lands (as Heroku is inactive already!).

PR #606, opened 2023-01-17
Fix ap

david-fisher:
Add the migration so the DB actually gets updated on deployment to production.
## Description

Add a migration
## Types of changes

## Checklist:

After completely deleting quepid, recloning, and adding the migration back in, I was able to successfully migrate the DB entry. Still unclear what this will mean in production, but I believe that the two steps listed for heroku:
heroku run bin/rake db:migrate -a quepid-staging
heroku restart -a quepid-staging
will need to be run after merging the pull request.


epugh (approved):
LGTM..   I do like some text in a migration about what is going on, (i.e why we are rerunning waht appears to be laoded by the db:seed), but that's just a nicety.

PR #709, opened 2023-04-30
Using Docker multi-stage build for jupyter-lite

mkr:
Small build change to build the jupyter-lite webapp only once and separate it from the build of Quepid container itself.


epugh:
we do need to keep the .gz file for deployments on heroku...   when you fire up rails on Heroku, we unpack the gz file...

PR #788, opened 2023-08-17
Refactor search engine into own endpoint

epugh:
## Description
Every try contains all of the data about the search engine.   This introduces an actual `search_endpoint` object that you can pick from.

## Motivation and Context
Every try drags along a copy of all the search engine end points.   We attempt to give them names, like `ES 1` or `Solr 2` but that's buggy, and not very meaningful to users.   Likewise, in a lot of places we show super long urls, like you might see on a cloud hosted version of your search engine, and htose are illegiable.   We want to give names to them that are human friendly!

Lastly, in lots of places an Admin type will set up the connection for use by the relevance folks, and maybe we want to introduce some usernames etc?   Also give us room to add NEW types of search engines with specific needs.

## Types of changes

## Checklist:


epugh:
Reminder: check and see if we can remove `searchEngineName.js` since this may all be handled on backend!

epugh:
- Need to test that we assign a search endpoint to a team when a case is shared with a team in the migration.

PR #826, opened 2023-09-11
PoC for Vectara integration

mkr:
This brings in a simple integration into the [Vectara](https://www.vectara.com) search engine.

This requires the [respective PR in splainer-search](https://github.com/o19s/splainer-search/pull/128) to be merged.

Currently, this is mainly supporting sending queries and getting documents back from Vectara.

Some limitations:
* There is currently no "explain" output.
* Vectara is separating `responses` from `documents` (where there can be multiple response for a single document). We are currently extracting the documents from the response.
* probably more...


app/assets/javascripts/services/docCacheSvc.js
```diff
@@ -55,7 +55,7 @@ angular.module('QuepidApp')
         var docIds    = Object.keys(docsToFetch);
         var resolver  = docResolverSvc.createResolver(docIds, settings, 15);
 
-        if ( docIds.length > 0 ) {
+        if ( docIds.length > 0 && settings.searchEngine !== 'vectara') { // 'vectara' does not support doc lookup by ID
```
epugh, line 58: This is totally fine...  And pionts to, in the future, we need some sort of "capablities" so you can be like `settings.searchEngine.supportDocLookup` instead ;-)

app/assets/javascripts/services/settingsSvc.js
```diff
@@ -83,6 +81,7 @@ angular.module('QuepidApp')
 
           escapeQuery:       true,
           apiMethod:         'POST',
+          customHeaders:     '',
```
epugh, line 84: why add it here if we remove it above?  In some places we do a check for either for existence of customheaders or an actual check of if it is `''`, are we maybe missing that someplace?

app/assets/javascripts/services/settingsSvc.js
```diff
@@ -91,7 +90,39 @@ angular.module('QuepidApp')
...
+            '  ]',
+            '}'
+          ].join('\n'),
+
+          escapeQuery:       true,
+          apiMethod:         'POST',
           customHeaders:     '',
+          fieldSpec:         'id:id',
+          idField:           'id',
+          titleField:        'title',
```
epugh, line 120: the titleField and idField...  are those values "standard" in Vectara land?   Do we/they want to host a demo Vectara that we can document and use in the wizard, kind of like the TMDB dataset?    And add to the https://github.com/o19s/quepid/wiki/Troubleshooting-Opensearch-and-Quepid#demo-opensearch-indexes-with-the-tmdb-dataset page....

PR #883, opened 2023-11-11
Manage Query Options Better

epugh:
## Description
Now the json query options is turned out as json in the response!

image

## Types of changes

## Checklist:


epugh:
I think I'm done!   Check this line out @mkr  https://github.com/o19s/quepid/pull/883/files#diff-99826bf177a5ca680203979c9576cd10ceb3e0dcc6b809e8bc2c96dc15568927

Here is the "glue" between the query options in Quepid and then how we treat it in Splainer.

epugh:
I went ahead and pushed forward a bit on adding options at the Search Endpoint and Case level, and letting search endpoint override Case options...   I'd like to play with this a bit to see how it works..  I actually am somewhat wondering if the Case options ends up being useful or not..     It may be that if you have a "case level" option, well than you just do it as part of the query sand box ;-).   I talked to the Amin at Vectara, and found out that now you don't need the customerId in the query, if it's in the custom headers..   which did sort of eliminate some of the reason for this search endpoint level query options!

Issue #829, opened 2023-09-12
Updating a book should automatically update the cases associated with it.


david-fisher:
When I perform a set of judgments via the book-based judging interface, I have to manually refresh the ratings from the book when I examine the case. That makes it easy to fail to see a change after judging, due to omitting that step. I would like the cases to be updated automatically.

PR #906, opened 2023-12-06
Merry christmas david fisher!

epugh:
## Description

Use ActiveJob to register that we want to update any cases related to a specific judgement (via a book releationship) to have update ratings.

## Motivation and Context
See #829  for why.

This is also a "practice" round for understanding how well does ActiveJob work for us in Quepid-land.....  

## How Has This Been Tested?
Added a test.  Now using manually during rating to see...

## Types of changes

## Checklist:


app/assets/javascripts/controllers/scorer.js
```diff
@@ -18,20 +18,21 @@ angular.module('QuepidApp')
...
+      $scope.communalScorers    = [];     
       $scope.selectScorer       = selectScorer;
       $scope.usingDefaultScorer = usingDefaultScorer;
       $scope.communalScorersOnly = configurationSvc.isCommunalScorersOnly();
 
       if ($scope.activeScorer.scorerId && ($scope.activeScorer.scorerId !== 'default')) {
         $scope.activeScorer.scorerId = parseInt($scope.activeScorer.scorerId);
       }
-
+
```
david-fisher, line 29: Gratuitous whitespace change...

app/assets/javascripts/controllers/scorer.js
```diff
@@ -18,20 +18,21 @@ angular.module('QuepidApp')
...
         $scope.activeScorer.scorerId = parseInt($scope.activeScorer.scorerId);
       }
-
+     
       /*jslint latedef:false*/
       scorerSvc.list()
         .then(function() {
           $scope.userScorers     = scorerSvc.scorers;
           $scope.communalScorers = scorerSvc.communalScorers;
+           $scope.scorerAccessible = scorerAccessible();
```
david-fisher, line 35: nit: correct indendation

app/assets/templates/views/pick_scorer.html
```diff
@@ -3,6 +3,9 @@
   <h3 class="modal-title">How would you like to score this case?</h3>
 </div>
 <div class="modal-body">
+  <div ng-if="!scorerAccessible" class="alert alert-warning" role="alert">
+    The scorer {{activeScorer.name}} used by this case is NOT shared via a team with you, so if you switch away from it you won't have access to it again.
```
david-fisher, line 7: If I have access to a case that uses a scorer, and I want to see a different scorer's output, I surely ought to be able to return the case to the state that I found it, putting back the previous scorer, whether or not anyone has shared it.

epugh: i think that is kind of the question, how do you get access...  Right now, the only way to "get back access" would be to a) make you the owner, or b) Add the scorer to a team you are in...    We don't have any history type tracking that would let us pluck it out of the history.    The warning at least lets you know so you could ping the owner to share it with you?

epugh: i think the same pattern will apply if you have a try with a search endpoint, you can see it.  but if you switch it to one you either own or is in your team, well then you can't get it back....

david-fisher: Disagree. If I can see it, I can return to it. 

When I load up the case and view it, the case should record the displayed scorer, so that I can return to that state. More generally, each try should know not just its results, but the scorer at the time of the try. If I change scorers, I get a new try. Makes returning to alternate states easier (no queries were run to bring you this message...).

epugh: Humm...  that's an interesting point..   if you change score, you change try?   because then we *would* have the history.........   If we want to do that, let's go ahead and make a NEW ticket for just that feature.   Something like "YOu change anything, like a scorer, you get a new try"..  cause right now, I don't actually know what triggers the new try OTHER than rerunning the queries".   So, if you change the scorer, we just create a new try with that scorer...

epugh:
Ugh...  discovered, that if you create missing queries from a book, you lose the info need, or any qOptions..   I think a `query` object and a `query_doc_pair` need parity...

epugh:
Draft PR notes:

```
This one got away from me!  In this one PR we have some nice improvements in the lifecycle of books and cases:

1. Now when you judge a query doc pair for a book we update in the background any associated Cases.  This eliminates the need to manually refresh the case ratings from the book!   

1. Band-aid the awkwardness that Cases have Queries/Ratings while Books have QueryDocPairs + Judgements, and that there are attributes on a Query like information need, notes, or options that don't exist on a QueryDocPair.   So now we preserve them, which faciliates round tripping better.   Upload a Book?  Upload a Case?  It doesn't mattern ;-)  Long term we need to merge these two data structures...   

1. When books were first created, they didn't follow the same permissions/sharing structure that Scores and Cases did.  Now you can share a Book with multiple Teams!

1. And as an aside, anyone can edit a custom scorer, not just the Owner of it.  
```

Issue #848, opened 2023-10-17
Let an Admin post a message to users


epugh:
**Is your feature request related to a problem? Please describe.**
Sometimes we need to communicate to users that things are going to happen, and no way to communicate it.

**Describe the solution you'd like**
Be able to put up a bit o' HTML that is show to the users.  Let them click close, and not see it again.  If a new message is posted, then they will get the new message.   If the message is "timed" well, you remove the message, and then it no longer shows up.   It doesn't "expire" by itself.

**Describe alternatives you've considered**
We have "sort" of used intercom on the hosted solution.

**Additional context**

epugh:
Here is some potential HTML that could go in a messages box...

```
<div class="alert alert-info" role="alert">
  A simple info alert—check it out!
</div>
```

flaxsearch:
Some examples of the sort of messages I'd like to be able to put up (excuse my HTML):

<div class="alert alert-info" role="alert">
  Want to learn how to use Quepid to tune relevance? Join OSC's <a href=https://opensourceconnections.com/training/>self-led training courses</a>
</div>

<div class="alert alert-info" role="alert">
We've just added <strong>brand new features</strong> to Quepid - check out Eric Pugh's <a href=https://opensourceconnections.com/blog/>blog</a> for more details
</div>

<div class="alert alert-info" role="alert">
The Call for Papers for <i>Haystack US 2023</i> is <a href=https://opensourceconnections.com/cfp>now open</a> - submit a talk on how you use Quepid!
</div>

epugh:
This has been done for a while!

PR #911, opened 2023-12-21
Allowing admins to create "app announcements" which are then shown to users until they are marked as viewed

mkr:
First draft steps at implementing #848


epugh:
I am investigating the "right" way to do Ajax call back to dismiss the message....

epugh:
https://buttondown.email/bhumi/archive/how-did-we-go-from-ajax-to-turbolinks-to-hotwire/?utm_source=hotwireweekly&utm_medium=email&utm_campaign=week-50-turbo-8-deep-dive-hotwire-history-turbo-native-apps

epugh:
i told @flaxsearch  we would get this over the finish line this week.  I'm going to try for somethign simpler which is to show the alert message in line with the page load, and in that process flag it as "viewed" so we don't ahve to worry about the ajax call back......

epugh:
@mkr  could you take a spin through this?   And we'll get @flaxsearch blessing tomorrow and then can ship it.

PR #942, opened 2024-02-04
Reduce logging when creating snapshots

reid-rigo:
## Description

Snapshot creation is slow and clunky to test because the amount of logging it generates. This does two things:
- Filters snapshot.docs to not output the giant POST body
- Sets up TruncatingFormatter in dev mode to shorten SQL logging

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -11,12 +11,8 @@ Rails.application.config.filter_parameters += [
 ]
 
 Rails.application.config.filter_parameters += [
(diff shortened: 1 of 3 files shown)


config/initializers/filter_parameter_logging.rb
epugh, line 1: Definitly cleaner!

lib/truncating_formatter.rb
epugh, line 1: I tried for a while to make something work and was constantly foiled, this looks so elegantly simple!

epugh (approved):
Wonderful

PR #952, opened 2024-02-13
Check for max query text length

epugh:
## Motivation and Context
This caught us recently with a very long "question" in a RAG situation..

We may want to lengthen the query...

## Types of changes

## Checklist:


epugh:
@mkr  do you think we should just make queries longer than 500?   In your looking at logged queries was it relatively common?

mkr:
@epugh It was very rare. I would stick with 500 for now.

PR #954, opened 2024-02-16
Improve performance of ratings in JSON

reid-rigo:
## Description
`Case.includes(queries: :ratings)` is a heavy operation for large cases. 

This optimizes JSON building with a method (`Case#doc_ratings_by_query`) to load just the needed ratings columns in one shot.

### DIFF
--- a/db/migrate/20240214020519_update_query_updated_at_with_ratings.rb
+++ b/db/migrate/20240214020519_update_query_updated_at_with_ratings.rb
@@ -1,0 +1,7 @@
+class UpdateQueryUpdatedAtWithRatings < ActiveRecord::Migration[7.1]
+  def change
+    execute <<-SQL
+      UPDATE queries SET updated_at = GREATEST(updated_at, (SELECT MAX(updated_at) FROM ratings WHERE query_id = queries.id))
+    SQL
+  end
+end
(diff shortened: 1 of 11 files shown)


db/migrate/20240214020519_update_query_updated_at_with_ratings.rb
```diff
@@ -0,0 +1,7 @@
+class UpdateQueryUpdatedAtWithRatings < ActiveRecord::Migration[7.1]
+  def change
+    execute <<-SQL
+      UPDATE queries SET updated_at = GREATEST(updated_at, (SELECT MAX(updated_at) FROM ratings WHERE query_id = queries.id))
```
reid-rigo, line 4: Alternatively a new column could be added like `ratings_updated_at`

epugh:
How about we go through this next Ruby Code Jam???   It hink it could be great!

epugh:
@reid-rigo could we jump on a zoom and go through this..  I think my understanding has finally caught up enough over the past year to have a reaosnable conversation about this PR!   We recently moved a lot of stuff to allow Quepid to have a case with 5000 queries (previosu feasible limit was like 300 or 500) and so I suspect this could help with some of the remaining bottle necks.

reid-rigo:
@epugh yeah let's chat about it. I'll see if I can refresh my memory

PR #955, opened 2024-02-16
Add ForUserScope concern to DRY & simplify code

reid-rigo:
## Description
Most of these scopes are the same and only the `for_user` combined scope is used


app/models/team.rb
```diff
@@ -55,12 +55,4 @@ class Team < ApplicationRecord
...
             uniqueness: true
   # rubocop:enable Rails/UniqueValidationWithoutIndex
-
-  # Scopes
-  scope :for_user, ->(user) {
-    joins('
-      LEFT OUTER JOIN teams_members on teams_members.team_id = teams.id
-      LEFT OUTER JOIN users on users.id = teams_members.member_id
-    ').where('`teams_members`.`member_id` = ?', user.id)
-  }
```
reid-rigo, line 65: This appears unused

epugh: Could be!

epugh:
This is very exciting!   Will this reduce the number of queries we run?   I believe right now we run three, one for the directly owned, one for the teams, then we pluck the id's, and then do a query for that!

epugh:
the test #create#test_alerts_when_a_scorer_associated_with_a_book_does_not_exist failing is my bad!

test/models/case_test.rb
epugh, line 1: i notice only a test on `case_test.rb` is this becasue it would be basically the same darn test and reducndent for book and scorer?

app/models/concerns/for_user_scope.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module ForUserScope
+  extend ActiveSupport::Concern
+
+  included do
```
epugh, line 6: This is so pretty!

test/models/case_test.rb
reid-rigo: Yeah actually let me move these cases into a test file just for this concern, for clarity. Also, I'll add test cases for `Scorer.for_user`

epugh:
THank you fo rthis, I've pushed it out to app.quepid.com...   It "feels" a bit faster...    The homepage on app.quepid.com still struggles to load quickly sigh.

Issue #979, opened 2024-03-13
Duplicate Cases being returned


epugh:
**Describe the bug**
We get duplicate cases being returned.

Here is an example from the `bin/setup_docker` script:

![image](https://github.com/o19s/quepid/assets/22395/20e6fb9a-0f6e-4ae2-a4a0-fd584c8807b6)

**To Reproduce**

Go to home page!

Maybe from #955...

epugh:
FIxed in #980

PR #982, opened 2024-03-14
Use Rails native AR patterns to control the user scope

epugh:
## Description
Go back to using more native rails code for #979 .....

@reid-rigo can you give this a quick once over?  

## Types of changes

## Checklist:

### DIFF
--- a/app/controllers/api/v1/cases_controller.rb
+++ b/app/controllers/api/v1/cases_controller.rb
@@ -62,6 +62,7 @@ module Api
                      current_user.cases_involved_with.not_archived.with_counts.preload(:tries, :teams,
                                                                                        :cases_teams)
                        .left_outer_joins(:metadata)
+                       .select('cases.*, case_metadata.last_viewed_at')
                        .order(Arel.sql('`case_metadata`.`last_viewed_at` DESC, `cases`.`updated_at` DESC'))
                    end
         end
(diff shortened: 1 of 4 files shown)


reid-rigo:
Looks good. I think adding a test case makes sense here. This one catches the issue:

```ruby
it 'returns distinct cases' do
  case_ids = Case.for_user(users(:random)).pluck(:id)
  assert_equal case_ids, case_ids.uniq
end
```

epugh:
> Looks good. I think adding a test case makes sense here. This one catches the issue:
> 
> ```ruby
> it 'returns distinct cases' do
>   case_ids = Case.for_user(users(:random)).pluck(:id)
>   assert_equal case_ids, case_ids.uniq
> end
> ```

Nice simple test case!  I tend to a lot of setup and verbosity that isn't actually needed!

app/controllers/api/v1/cases_controller.rb
reid-rigo, line 1: It looks like all of these `else` cases need the `select`, right? Since they're all looking at `last_viewed_at`

epugh: Thansk for the check, this led me to wondering who used the method and discovering that the code was added on the backend, but we never quite the front end JS to properly work...   So removed the code.

PR #976, opened 2024-03-08
Imagine if we could run 1000's of queries nightly in Quepid!!!

epugh:
## Description
This takes the existing Angular1 app and makes a small stripped version that just runs the query and the splainer stuff, and dumps out a text field in the "Snapshot" JSON format.   Then, this is exposed by the api `/api/agent_q/fetch` as the JSON response.    Oh, how does the API get the data?   Glad you asked _friend_!   The API instantiates Chrome and uses Chrome to invoke the stripped down case!   It's like that movie Inception ;-).    

```mermaid
sequenceDiagram
    Alice->>+QuepidAPI: I want to run a query for case 5 and get back the docs.  /api/cases/5/agent_q/
    QuepidAPI->>+AgentQ: Can you run a query for me.
    AgentQ->>+Chrome: Can you run a query for me.
    Chrome-->>-QuepidAPI: Please run for me /core/5/stripped
    QuepidAPI-->>-Chrome: Here is the HTML page with the data you want.
    Chrome->>AgentQ: Here is the rendered HTML page
    AgentQ-->>QuepidAPI: Here is the json response you want
    QuepidAPI-->>Alice: Here is the query doc data in JSON
```

## Motivation and Context
We have lots of logic in JavaScript and to move it out of there appears to be sissyphean...   Leading us down the "big rewrite" path.    Let's encapsulate the Angular1 and not expose it to more things.

Let's make it possible to run lots of queries...   So if you have 5000 queries, let's run them in batches of 50, and we invoke the browser 100 times.   Or batches of 10 and invoke it 500 times!

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)
- [] Improvement (non-breaking change which improves existing functionality)
- [] New feature (non-breaking change which adds new functionality)
- [] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [] My code follows the code style of this project.
- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.
- [] I have read the **CONTRIBUTING** document.
- [] I have added tests to cover my changes.
- [] All new and existing tests passed.

PR #1039, opened 2024-06-24
Spike 5000 queries round deux

epugh:
## Description
Demonstrate running 5000 queries in a background process, storing the results in a Snapshot.

Ideally, we run the 5000 queries, building the snapshot.   We then make the front end interact as if it was a CSV Upload of data.   If that works, great.  If it still doesn't scale, then we have a seperate UI.  And we figure out how to run not just A scorer, but lots of scorers, and store them..

- [x] Think about moving response_body to it's own table, it's really not part of the snapshot_queries and it appears to bloat the table a lot..   
- [x] Think about writing some code to convert the raw data from Solr into what Quepid expects.  Snapshot_queries, snapshot_docs, and some scores?
- [x] Add to snapshot picker a sense of date of the snapshot.
- [x] Convert Solr response in snapshot_queries into snapshot_docs.
- [x] Add logic to calculate a metric using a scorer.  Can we write a scorer in Ruby?
- [x] Read in scorer from database table, not file!
- [x] calculate a global score !
- [ ] Add a dashboard for all the case scorers?
- [ ] Experiment with loading a Snapshot in place of the actual running of the case.
- [x] Need to add the `number_of_rows` limit to reqeusting docs
- [x] Need to add the fl parameter to limit the fields being returned to what is defined in the case.
- [x] Tested with scorers that need access to the fields of a doc
- [ ] Need to be able to manage case scores, like deleting them if you have an odd ball one.
- [ ] Follow pattern in book and case of having a column to track the job.  So either add it to case.nightly_run_job or add it to the snapshot....
- [ ] Always show the most recent snapshot.   If there are no snapshots then prompt to run the queries with a big ass button in the middle.
- [x] Only keep most recent run;'s web_requests...  otherwise too much data.

Down the road, when we want multiple scorers:
- [ ] establish a single north star metric..  it shows up on the home page.
- [ ] establish a relationship between a `case_scores` and `scorers`, right now when you change scorers, we don't track that.
- [ ] Filter the graphs to only show scores that match the scorer being used.

## Motivation and Context
We need moar data!  This replaces an earlier attempt, https://github.com/o19s/quepid/pull/976 that kept the AngularJS app as the runner for Javascript.  It just didn't work...   Headless app wasn't the right way to go.


epugh:
Think about moving response_body to it's own table, it's really not part of the snapshot_queries and it appears to bloat the table a lot..   

Think about writing some code to convert the raw data from Solr into what Quepid expects.  Snapshot_queries, snapshot_docs, and some scores?

app/services/fetch_service.rb
```diff
@@ -0,0 +1,451 @@
...
+      #  { id: 1, value: 10, rating: 3 },
+      #  { id: 2, value: 20, rating: 0 }
+      # ]
+
+      # Calculate score with options
+      begin
+        scorer = snapshot.scorer
+        code = scorer.code
+
+        score = javascript_scorer.score(docs, best_docs, code)
```
epugh, line 161: @david-fisher we sometimes get NaN back from a scorer.  Is this okay?   Right now the logic is to force that to a zero.  That is also how the Angular app works too...   Or, do we fix it now?

app/services/fetch_service.rb
```diff
@@ -0,0 +1,451 @@
...
+        end
+      rescue JavascriptScorer::ScoreError => e
+        puts "Scoring failed: #{e.message}"
+      end
+
+      queries_detail[snapshot_query.query_id] =
+        { score: snapshot_query.score, text: snapshot_query.query.query_text }
+    end
+
+    # Opportunity here to fix the averageing logic
```
epugh, line 193: @david-fisher isn't blindly averaging for the case score NOT the correct thing to do, but it's what we have in Quepid today.   Should we change this up?   Or maybe I am thinking of something else?

test/services/javascript_scorer_test.rb
```diff
@@ -0,0 +1,216 @@
...
+
+      score = javascript_scorer.score(docs, best_docs, scorer_code)
+      assert_equal 0.33, score
+    end
+  end
+
+  describe 'recency scorer' do
+    let(:the_case) { cases(:case_without_score) }
+
+    test 'runs simple and tests eachDoc w/ a function' do
```
epugh, line 186: I'd love to get a bit more detail on how to pick a baseDate..  and if the baseDate could be "today" instead?  @david-fisher

app/services/fetch_service.rb
david-fisher: 0 for NaN is ok. Would like to know which scorer is giving the NaN's. Each individual scorer should probably ensure that it returns a meaningful value, rather than NaN.

app/services/fetch_service.rb
david-fisher: Each scorer should have a feature flag for sum vs average. As pretty much **all** of the scorers except DCG should be averaged, it is also ok to leave it averaging.

test/services/javascript_scorer_test.rb
david-fisher: base date of today is just fine. The one in the code was the date of the frozen in time index. You can use:

`baseDate = Date().now()`

epugh:
Maybe done?

PR #1049, opened 2024-08-07
Fix login and redirect problem with Keycloak

drieschel:
This PR fixes a problem when logging into quepid with Keycloak. 

## Description

Reverted `same_site: :strict` to `same_site: :lax`, which was introduced in v7.17.1.

## Motivation and Context

With Firefox it was not possible to login into quepid and in Chrome the redirect after authentication in Keycloak did not work properly.

## How Has This Been Tested?

Applied the fix in quepid instance and logging in over Keycloak worked again with Firefox as well as redirection in Chrome after successful authentication.

## Types of changes

## Checklist:

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 
 # Be sure to restart your server when you modify this file.
-Rails.application.config.session_store :cookie_store, key: '_quepid_session', same_site: :strict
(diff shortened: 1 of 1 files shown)


epugh:
thank you, I think I have been experiencing this elsewhere...!

epugh:
BTW, any thoughts on how to be able to restore the `:strict` or is it just not for us?

drieschel:
I think it is not possible, because the Keycloak authentication form usually resides under a different domain. Just guessing, didn't deep dive into that topic.

PR #1094, opened 2024-11-05
Move towards import maps

epugh:
This is a spike to see if we can load the AngularJS via import maps..   It still uses sprockets to handle making the public/assets/myfile-timestamped.js.   Presumably propshaft will take that over!


epugh:
We managed to eliminate sprockets in favour of propshaft in #1541.

PR #1113, opened 2024-11-29
Strip out SASS

epugh:
## Description
sassc-rails and sass-rails are abandoned.   Dart Sass is the future.  However, digging into it, we don't actually embrace sass, everything we do is just css...   

Let's remove the Sass as much as possible.

One thing is that for some reason sprockets-rails and sprockets autoloads sassc, so we can't totally remove the dependency.

## Types of changes

## Checklist:


epugh:
I think I am done...   So, Sprockets 4 will always invoke sassc, and so if you don't have it, you get an exception.  So we are stuck with that.    Also, Mission Control Jobs 0.6.0 still embeds sass, and of course, it doesn't work with latest sassc, only with dartsass.    And Sprockets4 doesn't support dartsass...   so kind of at an impass..    I think to move forward, we need to embrace jssbundling/cssbundling, or fork sprockets 4?

PR #1127, opened 2024-12-06
Restore proper support for notifying users about Cookies being required policy

epugh:
The cookie logic no longer worked, and was only run on the cases page.  The links to the cookie policy didn't render either.

## Description
The popup for saying we need cookie support would happen, but never be stored.  The logic only ran on the core pages, not the homepage.   So you just got the popup over and over and over.

## Motivation and Context
Be less annonying!  And have it show up on the logical page, the homepage!


epugh:
Don't forget to make it all conditional on the cookies url being populated.  Update the docs too!

PR #1167, opened 2025-01-10
Review database structures

epugh:
## Description
Review the database structures for best practices

## Motivation and Context
Found that we never added indexes to search end points, and it slows down queries a LOT.  What else are we missing?


epugh:
This is going to be a multi step process till we can go through the ones we care about, and then hopefully remove the gem ;-)

PR #1185, opened 2025-01-18
Remove permissions architecture in favour of roles

epugh:
## Description
This strips out the rich permissions structure that we've had for years in favour of using the "are you an Administrator or not" approach.

## Motivation and Context
We thought having really fine grained permissions would be important to Quepid, but what we've seen is that it doesn't matter.  Indeed, if anything, the permissions structure is confusing to users.   Additionally, we've mostly cared "are you an Admin?" to decide if certain things can be done or not.

With Quepid8 coming out soon, this is a nice time to refactor and strip out this code.

In the future, we may do more with roles.  For example, a `judge` might see certain things, or not see other things.   

## How Has This Been Tested?
ran the tests and manually.


epugh:
We need to do some manual testing, make sure everything still works as expected..  I hit the high points...

PR #1187, opened 2025-01-18
Properly nest returning judgements for a book under the book route

epugh:
## Description
Move judgements.csv from /book/:id.csv to /book/:id/judgements.csv.  

This matches up with the existing /book/:id/judgements.json API end point!

## Motivation and Context
Be restful!   Don't return book details in json and judgement details in csv at the same url!

Will need to fix Jupyterlite notebook to follow this pattern.


epugh:
@david-fisher FYI, this was something I noticed when you fixed one of the notebooks, that the path to get the CSV was conceptually wrong.  This PR fixes that pathing, and will need an update to the notebook ;-(.

PR #1258, opened 2025-03-03
make llm service and options configurable from environnment variables

frutik:
## Description
configuration of llm service externalized into env variables. added configurable timeout for interactions with llm service

## Motivation and Context
in current implementation all details of llm, except the api key, are hardcoded. proposed chages allows specifying as llm service any llm compatible with openai completion api (industry standard)

## How Has This Been Tested?
tested locally on llama2 / ollama runtime - https://github.com/frutik/quepid-ollama/tree/master

## Types of changes
-  Improvement (non-breaking change which improves existing functionality)

## Checklist:

### DIFF
--- a/app/services/llm_service.rb
+++ b/app/services/llm_service.rb
@@ -48,7 +48,16 @@ class LlmService
   # rubocop:disable Metrics/MethodLength
   # rubocop:disable Metrics/AbcSize
   def get_llm_response user_prompt, system_prompt
-    conn = Faraday.new(url: 'https://api.openai.com') do |f|
+    llm_service_url = ENV['LLM_SERVICE_URL'] || 'https://api.openai.com'
+    llm_model = ENV['LLM_MODEL'] || 'gpt-4o'
+    llm_key = ENV['LLM_KEY'] || @openai_key
+    llm_timeout = ENV['LLM_TIMEOUT'] || 30
+    conn = Faraday.new(
+      url: llm_service_url,
+      request: {
+        timeout: llm_timeout,
+        open_timeout: llm_timeout
+    }   ) do |f|
       f.request :json
       f.response :json
       f.adapter Faraday.default_adapter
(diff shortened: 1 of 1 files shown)


epugh:
This is really cool...    I'll have to mull a little on this, but it really points to how we can make the LLM more plugabble.    I am also keeping an eye on projects like https://github.com/activeagents/activeagent and https://github.com/patterns-ai-core/langchainrb to see if they get enough adoption to make sense to depend on..

epugh:
I looked a bit through this, and super excited about this.    

So, as we think about embarking on dropping ✨ emojies everywhere to indicate AI (or is it that diamond?) I am realizing we need some good abstractions over the various providers.    In the Rails world I have seen ActiveAgent, Langchainrb, and others that purport to do this.   

My gut feeling is that we should NOT be too ambitious, and maybe just roll our own support for OpenAI and Llama (in the same vein of this PR and the #1256)...   

To get this PR in I think we need:

* Tests
* More robust configuration
* Some documentation in the https://quepid-docs.dev.o19s.com/ docs site...

frutik:
> the various providers. In the Rails world I have seen ActiveAgent, Langchainrb, and others that purport to do this.

Don't overcomplicate it with unnecessary dependencies. In the end, they all support some kind of HTTP API.

frutik:
> More robust configuration

Well, the 12-factor app manifesto states that the best configuration method is via environment variables. 🙂

So, indeed, this way of configuring is not very flexible (as it still allows only one LLM), but it does at least provide a choice of which one (from the list of compatible with openai api) to use.

epugh:
Couldn't figure out how to push to here, so I created https://github.com/o19s/quepid/pull/1275

epugh:
Thank you for opening this PR!   I couldn't have done the Ollama work without following the steps that you laid out in this PR and the Quepid-ollama repo!

PR #1295, opened 2025-03-29
Spike out using different approach for docs

epugh:
## Description
Trying out oas_rails gem

## Motivation and Context
apipie is a bit awkward to write docs for, and the UI is lacking.

I am getting an error trying to load "https://unpkg.com/rapidoc@9.3.8/dist/rapidoc-min.js", I get an Internal Server Error!


epugh:
cc @a-chacon, witht he new gem it is much vetter!

a-chacon:
> cc @a-chacon, witht he new gem it is much vetter!

I am happy you are using it! Hope it helps you.

I will release the next version with the class level tags support.

epugh:
@a-chacon, i have a question.  I see this:

image

Is the example being populated from my books.yml fixture?   the very first fixture is:

image

epugh:
> @a-chacon, i have a question. I see this:
> 
> image
> 
> Is the example being populated from my books.yml fixture? the very first fixture is:
> 
> image

Okay, I see that there was a drop down and different fixutres load...   I guess I need to update ther eferences to _id's however right?

a-chacon:
hey @epugh yes, it is. OasRails just read what is in the fixtures for now. And I know there could be differences and may not be so exactly as the request body should be. But currently I am motivated to work on this: https://github.com/a-chacon/oas_rails/pull/124

So I don't want to add more specific Rails logic for the moment, and fixtures are Rails specific.

I think the best way to do it is describing in the comments the request_body/responses and examples, but I am aware that is much work and repetitive task, the class levels tags can help for this, but I need another approach too for reuse structures. 

Something like:
```ruby
# @structure Article [Hash{id: String}]
class ArticlesController < ApplicationController
  # @response Success(200) [Array<Article>]
  def index
  end

  # @response Success(200) [Article]
  def create
  end
end
```

It is just an idea. But for the moment I will be focused on the PR I show before.

a-chacon:
It is a pleasure! I will continue enhancing OasRails!

Issue #1309, opened 2025-04-10
I want the latest snapshot without thinking about id's... Just give me the latest.


epugh:
**Is your feature request related to a problem? Please describe.**
The two step of looking up the case, getting all the snapshots, finding the latest, and then looking up that snapshot is a pain.   We already have https://quepid.herokuapp.com/api/cases/6789/snapshots/123.json

**Describe the solution you'd like**
I would like to hit https://quepid.herokuapp.com/api/cases/6789/snapshots/latest and get back the latest snapshot.

**Additional context**
Big improvement for minimal API coding!

cc @david-fisher @sjarmak

PR #1312, opened 2025-04-11
first pass of adding /case/ID/snapshots/latest

epugh:
This is a fix for #1309 which lets us not worry about specific snapshot ids..


epugh:
image

PR #1344, opened 2025-05-22
Attach image to OpenAI request

khrabrovart:
Added explicit image attachment for LLM Chat Completions API request.

## Description
Here I added an additional image attachment to the OpenAI request only if document contains the Image field.
Otherwise we just create a simple text prompt.

Basically we create the following structure according to this article: https://platform.openai.com/docs/guides/images-vision#analyze-images
```
curl https://api.openai.com/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{
    "model": "gpt-4.1-mini",
    "messages": [
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "What is in this image?"
          },
          {
            "type": "image_url",
            "image_url": {
              "url": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"
            }
          }
        ]
      }
    ],
    "max_tokens": 300
  }'
  ```

## Motivation and Context
Currently if document has the Image field containing image URL, this field is just passed to LLM in a text prompt making it impossible for most (if not all) of OpenAI models to recognize and analyze it. In our case we would like AI judge to recognize the image content before evaluating the score. I think this logic would be reasonable in general, not only in our case.

### DIFF
--- a/app/services/llm_service.rb
+++ b/app/services/llm_service.rb
@@ -37,19 +37,35 @@ class LlmService
     judgement
   end
 
+  # rubocop:disable Metrics/MethodLength
   def make_user_prompt query_doc_pair
     document_fields = query_doc_pair.document_fields
-    fields = document_fields.blank? ? '' : JSON.parse(document_fields).to_yaml
 
-    user_prompt = <<~TEXT
+    fields = if document_fields.blank?
+               {}
+             else
+               JSON.parse(document_fields)
+             end
+
+    text_prompt = <<~TEXT
       Query: #{query_doc_pair.query_text}
 
       doc1:
-        #{fields}
+        #{fields.to_yaml}
     TEXT
 
-    user_prompt
+    prompt = [
+      { type: 'text', text: text_prompt }
+    ]
+
+    if '' != fields['image'].to_s.strip
+      image_url = fields['image']
+      prompt << { type: 'image_url', image_url: { url: image_url } }
+    end
+
+    prompt
   end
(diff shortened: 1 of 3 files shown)


epugh:
This is a fun PR!   I'll take a look see at it probably tomorrow...    So, longer term, I want to get one of the more sophisticated tools (dare I say agentic??) integrated into Quepid so that instead of a bunch of hard coded rules, like "if there is an image, use it in the prompt", we could instead try to ask a question like "Of all the data we have here, which is some `document_fields`, of which some are images, and others are http links, figure out what would be best for answering this prompt".   You could imagine that the agent makes a decision like "oh, we pick certain images, or we go and download the html form the reference webpage, or we prioritize title over description, and give all that to the judge"

frutik:
There is a problem with the agents - they should be run somewhere.

charliehullmoonpig:
I think there's also value in explainability and being very sure what you're passing to the LLM or agent for judging. It should be as close to what a human judge sees when using Quepid, otherwise it might impact agreement (or more likely, validation of the AI judge).

epugh:
Totally fair point @charliehullmoonpig ...  We know that inter rater reliability is hard enough with out adding more variablity!

PR #1433, opened 2025-08-04
Revamp annotation tooling

epugh:
## Description
Introduce Rails Lens

## Motivation and Context
We have three gems: annotaterb, database_consistency, and rails-erd, and I think they can be replaced with rails_lens.

## Types of changes

## Checklist:


epugh:
Before I can finish this, I want to get rid of rails-erd, but I need to figure out how to render the mermaid style erd doc: https://github.com/seuros/rails_lens/issues/15

.rails-lens.yml
```diff
@@ -0,0 +1,8 @@
+# .rails-lens.yml (optional)
+adapter: auto  # auto-detects PostgreSQL/MySQL/SQLite3
```
seuros, line 2: i will remove adapter, marker_format. That a relic from my older version (when i was using only PG and mysql was not tested)

.rails-lens.yml
```diff
@@ -0,0 +1,8 @@
...
+marker_format: rails-lens:schema  # LLM-friendly markers
```
seuros, line 3: the marker_format will be removed in the next patch version, it better to have consistent across the ecosystem, than having every project creating their own confusing marker.


If you have a better naming, i'm all hears.

.rails-lens.yml
```diff
@@ -0,0 +1,8 @@
...
+include_notes: true  # Performance recommendations
+extensions:
+  enabled: true  # ClosureTree, PostGIS, etc.
```
seuros, line 6: While you don't use any of the extensions currently, you can build you own for your business requirements.

For now ClosureTree and PostGIS are embedded, but they will move to their respective gem before we hit 1.0.0.


Note: Fixture and Fabricators don't get annotated for now.

.rails-lens.yml
epugh: thanks

PR #1479, opened 2025-09-25
Dependency updates + for funsies, we eliminated our old D3 v3 dependency.   Oh, and fix annotation renderings in core scoring chart!

epugh:
This got a bit away from me...  Update Ruby, NPM nodes, Bundler, Node, and finally finally eliminated our weird dependency on D3.js that was V3 only..  It's now on the latest.


epugh:
Annotations are back!  
image

PR #1591, opened 2026-01-05
Lets fix query handling in mapper wizard, you can skip easily providing the query template

epugh:
## Description
Ensure the specific query parameter, i.e `q=#$query##` or whatever you need gets provided.

## Motivation and Context
Mike H found a bug!

## How Has This Been Tested?
Manual


epugh:
Query mapper fix and then I did what I tell others not to do, a unrelated change in the same pr.

PR #1597, opened 2026-01-08
Smarter averaging of ratings

epugh:
## Description

  New Rating Calculation Logic

  Previous approach: Simple average of all judgementss

  New approach (optimistic-pessimistic):
  1. Take the three highest ratings from all judgements (optimistic: trust the best judges)
  2. If those top 3 ratings agree (all same value), use that value
  3. If they disagree, use the minimum of the top 3 (pessimistic: trust the lower rating, assuming judges tend to overrate)
Written up at https://quepid-docs.dev.o19s.com/2/quepid/63/how-judgements-are-averaged-into-a-rating-in-a-case

  Files Changed

  app/services/ratings_manager.rb
  - Added new calculate_rating_from_judgements method implementing the business logic
  - Updated sync_judgements_to_ratings to use the new calculation method

  test/services/ratings_manager_test.rb
  - Added 9 new test cases covering:
    - Single judgement
    - Two judges agree/disagree
    - Three judges agree/disagree
    - More than 3 judgements with top 3 agreeing
    - More than 3 judgements with top 3 disagreeing
    - Lower ratings being ignored when determining consensus
    - Float (implicit) ratings

  Examples

  | Judgements      | Top 3     | Result | Reason                             |
  |-----------------|-----------|--------|------------------------------------|
  | [3, 3, 3]       | [3, 3, 3] | 3      | All agree                          |
  | [3, 2, 1]       | [3, 2, 1] | 1      | Disagree → min                     |
  | [3, 3, 3, 1, 0] | [3, 3, 3] | 3      | Top 3 agree, lower ratings ignored |
  | [3, 2, 2, 1, 0] | [3, 2, 2] | 2      | Top 3 disagree → min of top 3      |

## Motivation and Context
Cause we had stupid logic

## How Has This Been Tested?
@david-fisher can you review and bless?

### DIFF
--- a/test/jobs/update_case_ratings_job_test.rb
+++ b/test/jobs/update_case_ratings_job_test.rb
@@ -8,7 +8,7 @@ class UpdateCaseRatingsJobTest < ActiveJob::TestCase
   let(:case_with_book) { cases(:case_with_book) }
   let(:book) { books(:book_of_star_wars_judgements) }
 
-  test 'account is charged' do
+  test 'update case rating' do
     case_with_book.queries << Query.create(query_text: 'my search')
 
     query_doc_pair = QueryDocPair.create(query_text: 'my search', doc_id: 'DOC123456')


epugh:
@david-fisher I changed logic a bit so if there are 2 judgments, we average (which is what we had before).  and if 1 judgemnt, then we just use that.   

Thoughts?

Issue #1287, opened 2025-03-19
Add an OmniAuth Provider for Microsoft Entra ID


schuch:
**Is your feature request related to a problem? Please describe.**

We want to use Microsoft Entra ID for authentication of Quepid users.

**Describe the solution you'd like**

Quepid should package an Omniauth provider that allows integration with Microsoft Entra ID or more generally with OpenID Connect.

**Describe alternatives you've considered**

Possible candidates:

- https://github.com/pond/omniauth-entra-id
- https://github.com/omniauth/omniauth_openid_connect

**Additional context**

- https://learn.microsoft.com/en-us/entra/architecture/auth-oidc

epugh:
I think this is an awesome idea...   A BIG reason we went with the Omniauth library was the promise of extending auth to multiple systems.   It worked great for the Keycloak and Google Auth providers.

My one wrinkle on implementing this is that I don't have a Microsoft Entra ID set to test any code changes against...   When adding auth, I always struggle with the "and how do I know it actually works!".   Is there a Docker type setup that I could test against?  Or OpenID "test" system otu there that I could use with Quepid to know the auth flow works.

You mention OpenID Connect, I believe that Keycloak is an OpenID Connect based system?

JPrevost:
@epugh I found this ticket when considering opening an Issue to ask if there would be support for adding https://github.com/omniauth/omniauth_openid_connect and saw it noted here so thought I'd check in here rather than opening a new ticket.

We will very likely add support for https://github.com/omniauth/omniauth_openid_connect in our forked copy, and if we pursue that would be happy to do it in a way that could be merged back here if there is interest.

As for why we'd want this rather than the existing keycloak integration:

- https://github.com/omniauth/omniauth_openid_connect is part of the official Omniauth GitHub organization
- https://github.com/ccrockett/omniauth-keycloak is not part of a GitHub org at all
- Keycloak may be able to support OpenID Connect, but it is a separate product and it feels worthwhile supporting OpenID Connect directly rather than supporting it via a different product specific implementation. Maybe they are 1:1 feature wise, maybe not.
- And most importantly (to me, selfishly), we rely on https://github.com/omniauth/omniauth_openid_connect for all of our Rails applications and it works great :) 

We anticipate having to maintain a fork for other purposes for a bit at least, so no worries if there is not interest upstream, I just thought I'd check in because it will require careful consideration if we add the integration in a fully generic way versus a semi-generic way we'd introduce if doing it just for us.

[...]

PR #1690, opened 2026-04-10
OpenID Connect Support

JPrevost:
## Description

* OpenID Connect is a popular authentication protocol that allows users to log in using their existing accounts from other providers.
* Adding support for OpenID Connect will enhance the user experience by providing more login options.
* This feature is disabled by default, so sites not wishing to provide it are not affected.
* All existing authentication options remain unchanged.

How does this address that need:

* Implemented OpenID Connect support using the omniauth-openid-connect gem. This models the similar implementation in place in this codebase for Keycloak and Google.

## Motivation and Context

https://github.com/o19s/quepid/issues/1287

## How Has This Been Tested?

We are using this configuration against an Okta OpenId Connect provider in our local environment.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves existing functionality)

## Checklist:


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -44,6 +44,30 @@ def google_oauth2
...
+
+      if @user.errors.empty?
+        session[:current_user_id] = @user.id # this populates our session variable.
+
+        # This is added because it is included in the google_oauth2 and keycloakopenid flows, and we want to be
+        # consistent across all OIDC providers. It also ensures that new users get a case created for them, but the
+        # comments on the other flows imply that this is only for new users, which is not the case so it is not entirely
+        # clear what the intent is. If the intent is to only create a case for new users, then we should probably check
+        # if the user is new before creating a case.
+        @user.cases.build case_name: "Case #{@user.cases.size}"
```
epugh, line 61: You know what, I think this is leftover from the dark ages when we didn't have a "homepage" for Quepid.    You literally just got dropped right into the Case creation screen.   

I think we don't need this, and honestly, probably should remove them from the others as well.

Also, when we formalize the "Judge (Rater) user persona" they won't even ever see cases!

app/views/sessions/new.html.erb
```diff
@@ -22,7 +22,17 @@
...
             <%= image_tag "openid-logo.png", :class => "black-to-white" %> Sign in with OpenId
-            
+
+          <% end %>
+        </p>
+        <% end %>
+
+        <% if Rails.application.config.devise.omniauth_providers.include?(:openid_connect) %>
+        <% email_login_only = false %>
+        <!-- <a href="https://www.flaticon.com/free-icons/digital" title="digital icons">Digital icons created by Freepik - Flaticon</a> -->
```
epugh, line 32: ???   Do we need this for crediting the image origin?

.env.example
```diff
@@ -69,13 +69,29 @@ EMAIL_SENDER=
 # This lets you disable the sorting if you experience the bug.
 QUERY_LIST_SORTABLE=true
 
-# OAuth Settings
+# OAuth Settings: remove these to disable Google OAuth or set them to your own values
 GOOGLE_CLIENT_ID=your_google_client_id
```
epugh, line 73: i wonder if we need to comment out these settings so when users first fire it up, they get reasonable defaults?   I believe the setup script just copies `.env.example` to `.env`, so mayb we need to comment these?  The keycload one is there I guess becasue when you fire up quepid in dev mode we already have it.

epugh: maybe we need `GOOGLE_CLIENT_ID= #your_google_client_id` ???

epugh (commented):
This is looking really good...    I wonder, can we drop the Keycloak set up and use openid to connect wiht Keycloak?  Any suggestions on how to test this?

app/controllers/users/omniauth_callbacks_controller.rb
JPrevost: Gotcha, that makes sense. Would you prefer I remove it from this OpenID Connect configuration or leave it so you can clean them all up at the same time?

app/views/sessions/new.html.erb
JPrevost: I copied this from the other implementations. I figured someone made a decision to do that intentionally in the past so I just went with it. I'm happy to adjust or leave as-is based on your preference.

.env.example
JPrevost: Defaults are hard for sure and I don't have a sense how frequently people benefit from having the non-functional Google option showing. It _might_ be good so they know it is an option and just needs proper configuration, or it might be really frustrating. Personally I didn't really mind it the way it is and it didn't take long to figure out I needed to comment it out to make it go away.

JPrevost:
@epugh While a Keycload server _can_ provide OpenID Connect services, I'm not confident whether the integration in this app is using OpenID Connect or another supported Keycloak authentication flow.

I suspect, but have not confirmed, removing the direct Keycloak option from Quepid would be a breaking change that requires re-configuration at best so I'd suggest if you want to remove the option it would be best to deprecate it first with a timeline to removal just in case it is in active use in production someplace.

epugh:
Hi @JPrevost I wanted to give you an update that I have been working on how to use Keycloak with OpenID directly..   I ran into one hiccup which is our dev Keycloak runs on http, not https, and that is an issue.  I found a ugly workaround, and then have opened up https://github.com/omniauth/omniauth_openid_connect/pull/211.   

If we do get this PR in, then we can actually REMOVE the `omniauth-keycloak` gem.    I have not pushed up my local changes to your PR yet pending the other PR getting in.   

If this is holding you up, we could merge this PR, and then when the openid PR lands, then do the refactoring thiere.

epugh:
This is the local version of my refactor...  [eric_refactor.patch](https://github.com/user-attachments/files/26931959/eric_refactor.patch)

JPrevost:
@epugh We are not blocked at all with this. We're using a fork with our OpenID Connect support in place and will eventually retire the fork if it lands here. No rush and thanks for taking the time to consider how best this fits in the app longterm.

epugh:
are you on relevance slack #quepid  channel?

epugh:
Okay, I've beat my head on this for a long time, but I haven't been able to get the omniauth-openid gem to get updated.    I'm going to push a fix or two here, and call it good.

epugh:
I ran into merge conflicts, so opened up #1701...

PR #1691, opened 2026-04-17
Adopt splainer-search 3.x with Angular shim for wired services

davidshq:
## Description

Replace the v2 Angular package with splainer-search 3.0.0 from the splainer-rewrite branch (pinned commit for reproducible builds). The new library is vanilla ESM and no longer registers an Angular module.

- Add splainer_search_adapter.js: wire createWiredServices(createFetchClient()) once and register the same service names on o19s.splainer-search, plus SettingsValidatorFactory and isAbortError, so existing DI is unchanged.
- Point angular_app.js at the local adapter instead of importing splainer-search directly.
- Update yarn.lock; splainer-search no longer pulls Angular as a dependency.

## Motivation and Context
We want to move away from the AngularJS version of splainer-search as part of our move away from AngularJS within Quepid - this is a first step.

## How Has This Been Tested?
Tested by @epugh and myself, including running tests.

NA

## Types of changes

## Checklist:

### DIFF
--- a/test/services/fetch_service_test.rb
+++ b/test/services/fetch_service_test.rb
@@ -181,8 +181,9 @@ class FetchServiceTest < ActiveSupport::TestCase
       end
 
       fetch_service.delete_extra_web_requests acase
-      assert_not_nil acase.snapshots.first.snapshot_queries.first.web_request
-      assert_nil acase.snapshots.second.snapshot_queries.first.web_request
+      ordered_snapshots = acase.snapshots.order(created_at: :desc)
+      assert_not_nil ordered_snapshots.first.snapshot_queries.first.web_request
+      assert_nil ordered_snapshots.second.snapshot_queries.first.web_request
     end
   end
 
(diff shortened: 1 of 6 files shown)


davidshq:
@epugh - I need to do some more testing, but theoretically this swaps out splainer-search 2.x for 3.x. It passes all current tests fine.

epugh:
OMG, I think it just worked!

epugh:
I think this is no longer draft ;-)

test/services/fetch_service_test.rb
```diff
@@ -181,8 +181,9 @@ class FetchServiceTest < ActiveSupport::TestCase
       end
 
       fetch_service.delete_extra_web_requests acase
-      assert_not_nil acase.snapshots.first.snapshot_queries.first.web_request
-      assert_nil acase.snapshots.second.snapshot_queries.first.web_request
+      ordered_snapshots = acase.snapshots.order(created_at: :desc)
```
epugh, line 184: thanks for this...  I wonder if this is a linux versus mac type timing thing?

davidshq: It sounds like it often works but that it can break (like it did with CI) depending on how the data was inserted into the db

PR #1702, opened 2026-05-14
Use RubyLLM as general interface.  Add support for vision in models.

epugh:
## Description
Instead of hand rolling our integrations, lets lean on RubyLLM which has become a defacto standard for integrating with LLMs.

## Motivation and Context
Simpler code, less hand rolling.

## How Has This Been Tested?
Manually (openai and ollama).


epugh:
@mbertani-nhn I would love it if you could test this PR with your models.  I want to make sure we didn't regress anything.

PR #1726, opened 2026-06-21
Fixes #1725

### LINKED ISSUE #1725: Fix local Docker setup configuration and documentation
There are 4 bugs related to documentation and config files.

Bug 1: Background Worker Math Crash [queue.yml ]
The Problem: The background worker crashes on boot with a math/string ArgumentError.
The Cause: ENV.fetch("SOLID_QUEUE_POLLING", 0.1) parses as a string when loaded from Docker environments, breaking the .seconds calculation.
The Fix: Update the config file to explicitly cast it as a float: <%= ENV.fetch("SOLID_QUEUE_POLLING", 0.1).to_f %>

Bug 2: Nginx Proxy Port Mismatch (502 Bad Gateway)
The Problem: Nginx proxy runs, but accessing http://127.0.0.1/ throws a 502 error.
The Cause: In docker-compose.yml, the app exposes ${APP_PORT:-33000}:3000. The .env file does not set APP_PORT, so it defaults to 33000 on the host. However, the internal Nginx configuration expects the app to be on port 3000. Nginx gets a connection refused.
The Fix: Update docker-compose.yml to use the fallback ${APP_PORT:-3000}:3000 and add APP_PORT=3000 to the .env.example file.

Bug 3: The MySQL Race Condition
The Problem: Running bin/docker server immediately after wiping containers causes the Rails worker to crash with Can't connect to server on 'mysql'.
The Cause: Docker starts the app worker and MySQL simultaneously. MySQL takes 10+ seconds to initialize its socket. The worker tries to connect instantly, fails, and kills the process tree.
The Fix: Update the documentation to explicitly instruct users to run docker compose up -d mysql, wait 15 seconds, and then start the server. [Alternatively, they need to add a healthcheck dependency for MySQL in the local  docker-compose.yml].

Bug 4: Missing Initial Database Seeding
The Problem: Once the race condition is bypassed, the worker crashes again with Unknown database 'quepid_development'.
The Cause: The standard boot scripts do not initialize the database structure for first-time Docker users.
The Fix: Add docker compose run --rm app bin/rails db:setup to the standard onboarding documentation.

These are the changes to be made.

yanaix10:
This PR fixes #1725


DEVELOPER_GUIDE.md
```diff
@@ -96,8 +101,22 @@ Run the Bash based setup script to setup your Docker images:
...
+```bash
+docker compose run --rm app bin/rails db:setup
+```
+
+#### 4. Running the app
+
+**Note on First Boot:** Docker starts the app and MySQL simultaneously. MySQL takes about 10-15 seconds to initialize its socket. To prevent the background workers from crashing on your very first run, start the database first:
+
+```bash
+docker compose up -d mysql
```
epugh: Can we add a health check in docker compose for MySQL that would pause Quepid till MySQL is fully up???

yanaix10:
I have updated the PR based on the feedback. I reverted the manual 'wait 15 seconds' instruction from the documentation and implemented the automated code fix. I added a healthcheck to the mysql service and updated the app's depends_on block to use condition: service_healthy. I also set keycloak and ollama to wait for service_healthy to ensure a clean boot sequence.

yanaix10:
Hey @epugh, I also added a health check to Keycloak. I implemented a socket-based health check (/dev/tcp/...) to replace the dependency on missing `curl` utilities in the micro-image.

This resolves the previous dependency issues during setup by properly leveraging container orchestration. 

Since this updates the setup flow, where is the best place to include the 'Resetting the Environment' steps in DEVELOPER_GUIDE.md?

yanaix10:
hey @epugh are changes made appropriate ? or do I need to make other changes?

PR #1728, opened 2026-06-23
store up some small fixes

epugh:
Going to see if we can slim down the docker image some.


epugh:
inconclusive results..

PR #1729, opened 2026-06-25
Fix #1722 - Remove legacy TLS protocol redirect logic and UI warnings.

### LINKED ISSUE #1722: Redirects to HTTPS on quepidapp.com when redirecting to HTTP
**Describe the bug**
If you attempt to reload using HTTP so you can access a Solr endpoint, it redirects to the HTTPS URL. This only occurs when you have an SSL cert configured, so is not reproducible with default setup when developing locally.

**To Reproduce**
1. Go to quepidapp.com
2. Click on Cases
3. Select New Case
4. Add title
5. Select existing endpointing
6. Select a Solr endpoint
7. Click Reload with HTTP
8. You are redirected to the same page with HTTPS.

**Expected behavior**
Should reload using HTTP.

**Screenshots**

**Additional context**
Would be interesting to install SSL cert locally and see if this issue occurs. If it doesn't, then it is likely a server config issue with Heroku.

yanaix10:
Summary - 
Removes deprecated HTTP/TLS protocol mismatch checks and associated UI warning logic to resolve the empty warning box rendering issue.

Code Changes:
- mainCtrl.js : Removed the needToRedirectQuepidProtocol evaluation block. Cleaned up if/else blocks for proxy Requests.
- wizardModal.js : Stripped the checkTLSForSearchEngineUrl() invocation from the validation lifecycle.
- wizardModal_spec.js : Removed deprecated Jasmine test blocks asserting checkTLSForSearchEngineUrl behavior.

Frontend - removed the code blocks responsible for warnings.

Both frontend and backend tests passed.

Final UI Block : 
Screenshot from 2026-06-25 02-39-51


app/assets/javascripts/controllers/wizardModal.js
```diff
@@ -513,11 +509,7 @@ angular.module('QuepidApp')
       }
       
       function changeProxySetting () {
```
epugh: do we even need this method if it's blank?

app/assets/javascripts/controllers/wizardModal.js
```diff
@@ -533,30 +525,6 @@ angular.module('QuepidApp')
       }
 
 
-      function checkTLSForSearchEngineUrl () {
-        if ($scope.pendingWizardSettings.proxyRequests === true){
-          $scope.showTLSChangeWarning = false;
-        }
-        else {
-          $scope.showTLSChangeWarning = caseTryNavSvc.needToRedirectQuepidProtocol($scope.pendingWizardSettings.searchUrl);
```
epugh, line 541: What happens if you are running Quepid on http and you point to a https search engine?  What we want to do is show them the message that "you must use the Proxy option when Quepid and your Search Endpoint are not on the same TLS"...   Is that capablity still here?

app/assets/javascripts/services/caseTryNavSvc.js
```diff
@@ -80,63 +80,10 @@ angular.module('QuepidApp')
       // to HTTP in order to make calls to a Solr that is running in HTTP as well, otherwise
       // you get this "Mixed Content", which browsers block as a security issue.
       // https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content
-      this.needToRedirectQuepidProtocol = function (searchUrl) {
```
epugh, line 83: this method `needToRedirectQuepidProtocol` may now need to be `specifySearchProxyRequired` in any caller maybe?

app/assets/javascripts/services/caseTryNavSvc.js
```diff
@@ -80,63 +80,10 @@ angular.module('QuepidApp')
...
-          separator = '&';
-        }
-        
-        quepidUrlToSwitchTo = quepidUrlToSwitchTo + separator + 'protocolToSwitchTo=' + protocolToSwitchTo;
-        
-        return [quepidUrlToSwitchTo, protocolToSwitchTo];
-      };
       
       // Return the protocol Quepid is on
-      this.getQuepidProtocol = function () {
```
epugh, line 126: this is here to check at the javascript/browser level what Quepid is on.   To make sure our search endpoint and Quepid are on matching protocols!

app/assets/templates/views/devQueryParams.html
```diff
@@ -179,13 +179,6 @@ <h4>Tuning Knobs</h4>
             <p>
               <a ng-href="{{createSearchEndpointLink(settings.selectedTry.searchEndpointId)}}" target="_self">More</a>
             </p>
-            <p>
-              <span ng-show="showTLSChangeWarning">
```
epugh, line 183: Yeah, we need this..

app/assets/templates/views/wizardModal.html
```diff
@@ -78,20 +78,12 @@ <h2>What Search Endpoint do you want to connect to?</h2>
                 Custom Headers must be a valid JSON object
               </span>
             </div>
-            <div ng-show="showTLSChangeWarning" class="alert alert-warning">
-              <span>
-              You have specified a search endpoint url that is on a different protocol ( <code>{{protocolToSwitchTo}}</code> ) than Quepid is running,
-              and this triggers your browser to prevent communication with the search endpoint.
-              <br/>Option 1: Reload Quepid to run on the correct protocol.
```
epugh, line 85: I think we still need this, but there are no longer TWO options.  Instead we need to enable the proxy support.   I could imagine that you have some button here called *Enable Proxy* that would actually update the search end point right here...

spec/javascripts/angular/controllers/queryParams_spec.js
```diff
@@ -91,16 +91,4 @@ describe('Controller: QueryparamsCtrl', function () {
     expect(scope.settings.selectedTry.curatorVars[0].inQueryParams).toBeTruthy();
   });
 
-  it('handles changing TLS from http to https when you start on http', function () {
```
epugh, line 94: yeah, we don't need this, but we do need to make sure we have a test for the detecting the mismatch logic.

epugh (changes requested):
I think this is a good first start.   However (there is always a however!), I want to make sure we aren't just removing the TLS protocol redirect logic, but also going deeper in faciliating to the user the "enable proxy" option.

I will try and play with this PR tomorrow and see, but I think we probably need to enhance teh UI to help a user who has a mismtached http/https pairing to enable the proxy use on the Search Endpoint.

yanaix10:
Hey @epugh  I  messed up with the UI part a bit in the previous commit. Fixing it now.
Please confirm. Our aim is to remove the http reload option completely. and prompting users to check the Enable Proxy Requests button whenever there is a mismatch between Quepid and the search engine. Also it is very confusing about if the quepid  http works well es/os  and all sorts of combinations.  It is very confusing for me.

I have thought about following changes to be made.  
The controller Logic files are alright and just need a little change. I have to fix the UI part.
- rename the func  needToRedirectQuepidProtocol to  specifySearchProxyRequired. and add on required logic to the empty func  changeProxySetting().
- update the UI Box - when ever there is a mismatch between the quiped protocol and the search engine endpoint's protocol.  Clearly mention the protocol type for both each time and prompting the user to Enable Proxy Button.

Does this approach looks fine ? and any suggestions or corrections.

epugh:
> Hey @epugh I messed up with the UI part a bit in the previous commit. Fixing it now. Please confirm. Our aim is to remove the http reload option completely. and prompting users to check the Enable Proxy Requests button whenever there is a mismatch between Quepid and the search engine. Also it is very confusing about if the quepid http works well es/os and all sorts of combinations. It is very confusing for me.
> 
> I have thought about following changes to be made. The controller Logic files are alright and just need a little change. I have to fix the UI part.
> 
> * rename the func  needToRedirectQuepidProtocol to  specifySearchProxyRequired. and add on required logic to the empty func  changeProxySetting().
> * update the UI Box - when ever there is a mismatch between the quiped protocol and the search engine endpoint's protocol.  Clearly mention the protocol type for both each time and prompting the user to Enable Proxy Button.
> 
> Does this approach looks fine ? and any suggestions or corrections.

Yeah!  I think you are on the right path with ^^ notes!  Basically, anytime Quepid and the Search Endpoint are NOT on the same `http` or `https` url, then we need to tell the user to enable the proxy capablity.

yanaix10:
Hey @epugh  I have made changes to fix it. 

Changes - removed the UI BOX - for reload quepid with http protocol thing completely and the TLS protocol logic handling it.  
Added a UI which shows a warning box  - when protocol mismatch.
In the previous set  up  warning and the reload button was shown in both Create a new .. ..  Use an Existing search ... 
Sections. 

But in the current set up - if the user is dealing with create a new Search ..... - I have added the waring box prompting the user to check proxy Box.
But for Use an Existing Search Endpoint - It will not validate the user untill the search endpoint is not on the same protocol as quepid. 

My doubt can proxy check can be anabled from Use an Existing Search box as well ? 
or I can just prompt the user to create an Endpoint with the particular protocol by enable the proxy check first. 
and the valid URL logic is working well as previously for both the sections. 

Can we allow the user to  enable proxy check when in Use an Existing section does the work flow allows this or how quepid handles protocols stuff ? 

Also  from Deep Source JS - I get many fixes required with string parsing and const var and let what has done be done about that ?

Please review the changes. I tcan be a little cluttered this time but I have tried my best. Also practicing to write clean code as we move further. I had to deal with a lot up set up and config issues in between but they  have become clear thorugh the docs and it is working pretty fine.

app/assets/javascripts/services/caseTryNavSvc.js
```diff
@@ -6,182 +6,134 @@
...
+
+			this.getTryNo = function () {
+				return tryNo;
+			};
+
+			this.specifySearchProxyRequired = function (searchUrl) {
+				if (searchUrl) {
+					var quepidStartsWithHttps = $location.protocol() === 'https';
+					var searchEngineStartsWithHttps = searchUrl.startsWith('https');
+					return (quepidStartsWithHttps !== searchEngineStartsWithHttps);
```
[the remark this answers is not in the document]
davidshq: This I think answers @epugh earlier question about accessing https from http.

davidshq:
@naitikcodes15 - fwiw, I always run GitHub Copilot on the code when I make a branch. I find it provides useful insights. Sometimes it is off base, but in this case I think pretty much everything is worthwhile.

yanaix10:
@davidshq I think it should validate Copilot Issue as well as they seem to be valid and the code need changes as 
suggested so I should go ahead and fix them.

davidshq:
Yes, make sure it makes sense in the code, but from my review they all look good.

app/assets/javascripts/services/caseTryNavSvc.js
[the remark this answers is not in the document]
yanaix10: @davidshq Browser mixed-content blocking only applies when the page is HTTPS and the endpoint is HTTP; HTTP pages can still call HTTPS endpoints. -  is this true for all search engines and search API keys ?

But in the current set up - if the user is dealing with create a new Search ..... - I have added the waring box prompting the user to check proxy Box.
But for Use an Existing Search Endpoint - It will not validate the user untill the search endpoint is not on the same protocol as quepid.

My doubt can proxy check can be anabled from Use an Existing Search box as well ?
or I can just prompt the user to create an Endpoint with the particular protocol by enable the proxy check first.
and the valid URL logic is working well as previously for both the sections.

@epugh  If you could please clarify this and finalise the  required changes.

yanaix10:
hey @epugh @davidshq  

I have treated both options - as isolated : create a new search endpoint & Use an existing endpoint.
warning is shown in the box user is working in unlike the  reload protocol logic.

Screenshot from 2026-07-08 03-07-37

Screenshot from 2026-07-08 03-06-41

This commit sets up the core logic to implement this. we can adjust UI as recommended.
If I did not break any other core functionality which I think I did n't we are good to go.

Please review it and with a little more changes we will be able to get this done.

epugh:
thanks....  I will take a looksee in the next few days!

epugh:
Back from vacation, and giving this a work out with the goal to merge next week.

yanaix10:
hey @epugh  There is a bug ig currently.  a green box waring box rendering in both the section. away for a long time from it so can't name it actually. I will too get back to this after reaching college : 4th of august.

epugh:
> hey @epugh There is a bug ig currently. a green box waring box rendering in both the section. away for a long time from it so can't name it actually. I will too get back to this after reaching college : 4th of august.

Awesome....    Thanks for communicating and no worries!   Would love your eyes on this before we do a merge.

PR #1772, opened 2026-09-04
Pick the oldest match when a search endpoint lookup is ambiguous

frutik:
SearchEndpoint.find_or_initialize_for_user looks an endpoint up by four fields - search_engine, endpoint_url, api_method and proxy_requests - and those do not identify a row uniquely. The same URL can be registered more than once with different options or credentials, and the fixtures already contain such a pair:

    matches: 2
      id=247532087  opts={"corpusId" => 12345}
      id=732477833  opts=nil

`find_by` takes whichever of them the database returns first. No query said which that should be, so the answer is undefined - and it genuinely differs: MySQL returns the first, PostgreSQL the second.

The consequence is not cosmetic. Importing a case calls this to decide which search endpoint the imported case gets attached to, so an import could silently pick up an endpoint with different options or different credentials than the one it was exported from. Nothing was making that choice stable, on any adapter - MySQL just happened to be consistent enough that nobody noticed.

Taking the oldest match makes the choice deterministic, and is the row MySQL was already returning, so behaviour is unchanged there.

This surfaced as ExportImportCaseFlowTest failing on PostgreSQL with `NoMethodError: undefined method 'transform_values' for nil` - the endpoint it picked had no options. JSON handling was ruled out first: a hash round-trips through a json column identically on both adapters.

The new test was checked to be a real guard rather than one that passes either way - making the lookup take the newest match instead makes it fail.

Verification

Full suite, MySQL:  1251 tests, 4238 assertions, 0 failures, 0 errors.
Full suite, SQLite: 1251 tests, 4236 assertions, 0 failures, 0 errors.
RuboCop: 435 files, no offenses.

Claude-Session: https://claude.ai/code/session_01UAEu5HNhDxXkebPy1DQ6rb

## Types of changes

## Checklist:

### DIFF
--- a/app/models/search_endpoint.rb
+++ b/app/models/search_endpoint.rb
@@ -96,7 +96,12 @@ class SearchEndpoint < ApplicationRecord
     lookup_params = normalized.slice(*lookup_keys)
 
     if lookup_keys.all? { |key| lookup_params.key?(key) }
-      endpoint = user.search_endpoints_involved_with.find_by(lookup_params)
+      # These four fields do not identify an endpoint uniquely - the same URL
+      # can be registered more than once with different options or credentials.
+      # find_by would take whichever row the database happened to return first,
+      # which is not defined and does differ between adapters. Take the oldest
+      # match instead, so the same import always attaches the same endpoint.
+      endpoint = user.search_endpoints_involved_with.where(lookup_params).order(:id).first
       return endpoint if endpoint
     end
 
--- a/app/models/search_endpoint.rb
+++ b/app/models/search_endpoint.rb
@@ -99,7 +99,7 @@ class SearchEndpoint < ApplicationRecord
       # These four fields do not identify an endpoint uniquely - the same URL
       # can be registered more than once with different options or credentials.
       # find_by would take whichever row the database happened to return first,
-      # which is not defined and does differ between adapters. Take the oldest
+      # which is not defined and does differ between adapters. Take the lowest-id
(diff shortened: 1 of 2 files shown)


epugh:
i wonder if we need to rethink this a bit?  Why would we have multiple search endpoints..  I mean, I guess you can manually make duplicates.  but, should we maybe change teh UI up a bit?   Is auto picking the oldest actually good?  Or should we alert "Hey, you have two candidates, please pick one"?     Also, why isn't this an ongoing issue?

frutik:
> i wonder if we need to rethink this a bit? Why would we have multiple search endpoints.. I mean, I guess you can manually make duplicates. but, should we maybe change teh UI up a bit? Is auto picking the oldest actually good? Or should we alert "Hey, you have two candidates, please pick one"? Also, why isn't this an ongoing issue?

i think better threat issues separately. we first can unblock postgres and later think how to fix root cause of the issue in ui

@epugh

epugh:
> > i wonder if we need to rethink this a bit? Why would we have multiple search endpoints.. I mean, I guess you can manually make duplicates. but, should we maybe change teh UI up a bit? Is auto picking the oldest actually good? Or should we alert "Hey, you have two candidates, please pick one"? Also, why isn't this an ongoing issue?
> 
> i think better threat issues separately. we first can unblock postgres and later think how to fix root cause of the issue in ui
> 
> @epugh

that works...   it won't make anything worse!

PR #1777, opened 2026-09-08
Add PostgreSQL as a supported database adapter

frutik:
Quepid runs against MySQL, and against SQLite since #1763. This adds PostgreSQL: `DB_ADAPTER=postgresql` gets a container, a database.yml entry, a schema that loads, and a CI job that keeps it working.

What it took

Only two compatibility shims, in
config/initializers/postgresql_schema_compatibility.rb, both mirroring the existing sqlite3_schema_compatibility.rb:

  * `size: :medium` / `:long` are MySQL blob and text width hints. PostgreSQL's text and bytea are unbounded so the option is meaningless, but the adapter rejects unknown column options outright with `Unknown key: :size`.

  * Column-level `collation:` holds MySQL collation names, emitted as a literal COLLATE clause; PostgreSQL raises `collation "utf8mb4_bin" for encoding "UTF8" does not exist`. MySQL's naming convention carries the semantics: `_bin` means byte-for-byte case-sensitive, which is PostgreSQL's "C". There is no built-in case-insensitive collation to map the `_ci` names onto, so those are dropped and the column inherits the default. Making them behave as they do on MySQL needs citext or a nondeterministic ICU collation, which is a schema decision rather than a compatibility shim.

Three things that looked like they would need shims and do not: index `length: 191` prefix indexes are silently ignored, table-level `charset:` is swallowed by TableDefinition's keyword catch-all, and `pg` installs as a precompiled binary so neither Dockerfile needs libpq packages.

config/application.rb previously disabled dump_schema_after_migration for SQLite specifically. Since db/schema.rb is dumped from MySQL and carries options no other adapter can reproduce, that rule now covers any non-MySQL adapter rather than naming them one at a time.

What made it work

Making PostgreSQL pass was almost entirely application work, not adapter work, and every piece of it was a real defect that MySQL had been absorbing quietly:

  * SELECT DISTINCT over a table with a json column (PostgreSQL's json type has no equality operator) - #1768 and the follow-up covering three more sites.
  * Numeric columns compared with LIKE, and params passed straight into integer columns - #1767.
  * Email lookups relying on MySQL's case-insensitive collation - #1766.
  * Queries with no explicit ORDER BY, where adapters disagree about NULL placement and ties have no defined order at all.
  * A search endpoint lookup that could attach the wrong credentials on import.
  * A weighted random sample that silently flattened rather than raising.

None of those were PostgreSQL bugs. PostgreSQL just declined to paper over them.

Ordering

The CI job added here is red until those application fixes land, so this change belongs last in the sequence. It also depends on the annotate fix - CI runs bin/setup_docker, which runs db:migrate, which regenerates every model's schema annotation from the live connection unless that is guarded.

The job mirrors test-sqlite exactly. That is what turns "PostgreSQL happens to pass today" into something that stays true.

Verification

With the application fixes applied, all three adapters pass the full suite:

    MySQL       1269 tests, 4288 assertions, 0 failures, 0 errors
    SQLite      1269 tests, 4287 assertions, 0 failures, 0 errors
    PostgreSQL  1269 tests, 4286 assertions, 0 failures, 0 errors

This branch on its own carries no application fixes, so only MySQL and SQLite pass here: 1256 tests, 0 failures on both. RuboCop clean.

DEVELOPER_GUIDE.md documents the third adapter and one thing that cost real confusion: `bin/docker r bundle install` does not persist, because the wrapper uses --rm and gems live in the image. Adding an adapter gem requires `docker compose build app`.

Claude-Session: https://claude.ai/code/session_01UAEu5HNhDxXkebPy1DQ6rb

## Types of changes

## Checklist:


bin/setup_docker
```diff
@@ -37,11 +37,17 @@ bin/docker r bundle config set --local no-document true
...
-  bin/docker r bin/rake db:reset
-fi
+case "${DB_ADAPTER:-mysql2}" in
+  mysql2)
+    bin/docker r ./wait-for mysql:3306 --timeout=60 -- bin/rake db:reset
+    ;;
+  postgresql)
+    bin/docker r ./wait-for postgres:5432 --timeout=60 -- bin/rake db:reset
+    ;;
+  *)
```
epugh: i wonder if it makese sense to assume a database adapter instead of `*`?  Since now it's mysql2, postgresql, or sqlite?

DEVELOPER_GUIDE.md
```diff
@@ -87,7 +87,7 @@ This guide provides detailed instructions for developers who want to set up, run
...
 
-Quepid supports two database adapters: **MySQL** (the default for development and test) and **SQLite** (the production fallback when `DATABASE_URL` is unset, with no separate database server to install or run). Set `DB_ADAPTER=sqlite3` in your `.env` file (or export it before running `bin/docker`/`bin/rails` commands) to opt into SQLite for development and test; leave it unset to use MySQL. See [Choosing a database adapter](#choosing-a-database-adapter) below.
+Quepid supports **MySQL** (the default for development and test) and **SQLite** (the production fallback when `DATABASE_URL` is unset, with no separate database server to install or run). Set `DB_ADAPTER=sqlite3` in your `.env` file (or export it before running `bin/docker`/`bin/rails` commands) to opt into SQLite for development and test; leave it unset to use MySQL. **PostgreSQL** (`DB_ADAPTER=postgresql`) is also supported, served by the `postgres` container. See [Choosing a database adapter](#choosing-a-database-adapter) below.
```
epugh: i think this needs to be reworded.   SOmething like Quepid supports Mysql, Postgres, adn Sqlilite.   Mysql is used as the primary dev and test datbase.  In production, if Sqllite is the default if mysql or postgres aren't set up.   You can choose to run Postgres in production.

DEVELOPER_GUIDE.md
```diff
@@ -150,14 +150,16 @@ This approach lets you run Quepid directly on your machine without Docker. It pr
 
 3. **Yarn**: Install Yarn package manager.
 
-4. **Database**: By default Quepid uses MySQL, matching production - install MySQL 8.0+. If you'd rather avoid running a database server locally, set `DB_ADAPTER=sqlite3` in your `.env` file instead - the `sqlite3` gem is enough, no separate install needed.
+4. **Database**: By default Quepid uses MySQL, matching production - install MySQL 8.0+. If you'd rather avoid running a database server locally, set `DB_ADAPTER=sqlite3` in your `.env` file instead - the `sqlite3` gem is enough, no separate install needed. For PostgreSQL, set `DB_ADAPTER=postgresql` and install PostgreSQL 14+, or run just the `postgres` container.
```
epugh: not sure that MySQL still "matches produciton".  if anything, by default in prod i'ts sqllite, so I don't think we should even says anything about matching production.

DEVELOPER_GUIDE.md
```diff
@@ -911,12 +913,23 @@ This section covers common issues you might encounter during development and how
 
 Quepid defaults to **MySQL**, matching production. Set `DB_ADAPTER=sqlite3` in `.env` (or export it before running `bin/docker`/`bin/rails` commands) to use SQLite instead (`storage/development.sqlite3` / `storage/test.sqlite3`) - nothing to start up separately. Both are exercised in CI (`.github/workflows/test.yml`), and the production Docker image (`Dockerfile.prod`) carries the `sqlite3` gem too, so it can run against either adapter via `DATABASE_URL`.
 
+**Adding or changing an adapter gem requires an image rebuild.** `bin/docker r bundle install` runs in a `--rm` container and gems live in the image, not a volume, so the install is discarded when the container exits. Run `docker compose build app` instead.
```
epugh: how will we handle things in our docker image?   Seems like maybe we just install all the gems needed,  so no need to run bundle isntall...   bundle install is a shortcut that LLM's often use, but a person just runs bin/setup_docker and bin/docker s...

DEVELOPER_GUIDE.md
```diff
@@ -911,12 +913,23 @@ This section covers common issues you might encounter during development and how
...
+
+Two things to know:
+
+- `config/initializers/postgresql_schema_compatibility.rb` makes `db/schema.rb`'s MySQL-only column options loadable: it allows `size:` (a MySQL text/blob width hint that means nothing to PostgreSQL) and maps MySQL collation names, `_bin` to `"C"` and dropping the case-insensitive ones. **Columns that are case-insensitive on MySQL are therefore case-sensitive on PostgreSQL.** Application code no longer depends on that - email lookups and searches fold case in SQL - but keep it in mind when adding queries. Making the columns themselves behave as they do on MySQL would need `citext` or a nondeterministic ICU collation, which is a schema decision rather than a compatibility shim.
+- `SelectionStrategy`'s weighted sampling, and anything else needing `RANDOM()` or a natural log, goes through `AdapterFunctions` rather than branching inline. These functions are spelled differently per adapter and getting one wrong returns wrong numbers rather than raising, so add new ones there rather than at the call site.
+
 Known SQLite-specific behavior to be aware of:
```
epugh: Maybe need a ### SQLite header introduced?  And align the writing style between the postgres and the sqlite?

epugh (changes requested):
I think it's just wording chagnes that are really needed...

DEVELOPER_GUIDE.md
frutik: It could be our next initiative: slimming down the Docker image. For now lets focun on finding wht can be broken :-)

frutik:
> I think it's just wording chagnes that are really needed...

@epugh 

• Committed and pushed to postgres-spike, updating PR #1777.

  Commit: 62b5d820 — Address database adapter review comments.

frutik:
all done. all wordings chaged. I hope all

epugh:
thanks @frutik....    next release will be abig one..  Quepid support three database platforms, plus onboarding Vespa a full featured search engine, plus lots of enahncemnts and tweaks...

epugh:
testing now...

PR #1789, opened 2026-09-12
Add qdrant search engine

frutik:
## Description

  Adds **Qdrant** as a wizard tile on the existing "Search API" (mapper-based) endpoint type,
  the same path Vespa uses — no new endpoint type, no changes to the search pipeline.

  - **`app/models/mapper_based_search_engine.rb`** — a `qdrant` definition alongside `vespa`.
    Talks to the collection-scoped `POST /collections/<collection>/points/query`, authenticates
    with Qdrant's own `api-key` header rather than HTTP Basic, and is proxied server-side.
  - **`db/mapper_based_search_engines/qdrant.js`** — `numberOfResultsMapper`, `docsMapper` and
    `ratedDocsQueryParamsMapper`. The docs mapper is schema-agnostic: it maps Qdrant's
    `result.points[].id/score/payload` envelope, so which payload field displays as title/body
    is decided entirely by the case's `field_spec`. You can point a case at a different
    collection without touching the mapper.
  - **`docs/endpoints_qdrant.md`** — a companion to `endpoints_solr.md` / `endpoints_opensearch.md`:
    standing up a collection, indexing documents, running both BM25 and dense retrieval, wiring
    it into the wizard, and issuing Quepid a read-only key.
  - **`bin/docker`** — drive-by fix, unrelated to Qdrant: `run|r` passed `$@` unquoted, so

❯ show me the text here

  it into the wizard, and issuing Quepid a read-only key.
- **`bin/docker`** — drive-by fix, unrelated to Qdrant: `run|r` passed `$@` unquoted, so
  `bin/docker r bundle exec thor user:create -a admin@example.com "Admin User" supersecret`
  split the quoted argument. Now `"$@"`.

### Design decisions worth reviewing

- **The default query uses `qdrant/bm25`, not a dense model.** It is the one model Qdrant runs
  locally inside the cluster, so the tile works on self-hosted Qdrant as well as Qdrant Cloud,
  where every other model needs Cloud Inference. Swapping in a dense model is a one-line edit
  in the Query Sandbox.
- **`supports_pagination` is left false.** Qdrant's response envelope carries no total-hits
  field, so `numberOfResultsMapper` can only report the documents already on screen. Both
  "Peek at the next page of results" call sites gate on `numFound > docs.length`, which a
  self-reported count can never satisfy. Naming the pagination params would not fix that and
  would actively break search — splainer-search defaults them as **strings** (`'10'`/`'0'`),
  and Qdrant's API rejects `"limit": "10"` with a deserialization error. The page size is a
  real integer in `query_params` instead, editable in the Sandbox.
- **`search_url` ships empty.** There is no public Qdrant demo cluster to point at the way the
  Vespa tile points at the o19s demo tenant, so the tile asks for the user's own collection URL
  and `url_format` shows them the shape. Likewise `custom_headers` carries a placeholder key,
  not a credential.
- **`ratedDocsQueryParamsMapper` handles two different lookups.** With the case's id field at
  `id` it filters with `has_id` (Qdrant's point id is not a payload key, so a payload `match`
  filter cannot reach it); point it at a payload key and it switches to `match`/`any`. Point
  ids are an untagged integer-or-UUID union, so digits-only ids are converted back to numbers —
  sending `"42"` makes Qdrant try to parse it as a UUID and reject the request.

## Motivation and Context

Qdrant users currently have to hand-write a mapper to evaluate relevance in Quepid. This makes
it a one-click wizard tile like Solr, OpenSearch and Vespa, and gives Quepid a worked example
of a vector database on the mapper-based path — the BM25 default means it is useful for lexical
relevance work out of the box, not just dense retrieval.

## How Has This Been Tested?

Three new test files, 26 tests / 64 assertions, all passing via
`bin/docker r bundle exec rails test`:

- `test/lib/qdrant_mapper_test.rb` — executes the shipped mapper JS and asserts on its output:
  the payload spread, the point id surviving a payload that contains its own `id` key, both
  rated-docs filter shapes, the integer-vs-string id conversion, and that the default
  `query_params` and `test_query` parse as the JSON body `Try#searchapi_args` will POST.
- `test/models/mapper_based_search_engine_test.rb` — the definition loads its mapper from disk,
  every definition has a unique id and an existing mapper file, and the pagination/auth/URL
  decisions above are pinned.
- `test/controllers/api/v1/mapper_based_search_engines_controller_test.rb` — the tile is served
  by the API.

`rubocop` clean on all touched Ruby files.

**Verified against a live cluster ** — the mapper was developed
against a real Qdrant Cloud collection following `docs/endpoints_qdrant.md`, but it would be
worth someone reproducing that doc end to end before merge.
    `bin/docker r bundle exec thor user:create -a admin@example.com "Admin User" supersecret`
    split the quoted argument. Now `"$@"`.

  ### Design decisions worth reviewing

  - **The default query uses `qdrant/bm25`, not a dense model.** It is the one model Qdrant runs
    locally inside the cluster, so the tile works on self-hosted Qdrant as well as Qdrant Cloud,
    where every other model needs Cloud Inference. Swapping in a dense model is a one-line edit
    in the Query Sandbox.

## Types of changes

## Checklist:


epugh:
We are going to need to make this UI better soon!   
image

epugh:
I have spun up a review app at https://quepid-pr-1789.herokuapp.com/

anush008:
Thanks @frutik 👍

PR #1795, opened 2026-09-16
Port ruby/backend fixes from angular-cleanup-phase3 into the Stimulus-migration branch

davidshq:
## Summary
Cherry-picks the Ruby/backend-only work from `angular-cleanup-phase3` (which has its own, different Stimulus migration) onto this branch as fresh commits, skipping everything Angular/Stimulus-specific since this branch already did that migration independently.

- Harden case-404 guards (`ScoresController`, `DuplicateScoresController`, `Analytics::TriesVisualizationController`) so a missing/inaccessible case renders the 404 page instead of a JSON blob or crashing; dedupe `shared/_error_messages` DOM ids.
- Harden LLM judgement scoring: reject non-numeric LLM ratings instead of silently coercing to 0.0, validate against the book's configured scale, and augment the judge's system prompt with the book's real scale/labels (sanitized against prompt injection).
- Add `PATCH books/import` to append data to an existing book; harden `BookImporter` against mass-assignment (a crafted import file could previously reparent a query_doc_pair into another user's book).
- Fix `@ai_judges` being `nil` when `BooksController#create` re-renders `:new` on a validation failure.
- Add missing test coverage (`ApiKey` model had none) and fix a manual-raise-vs-validation smell there; fix a stray "Websocker" typo.
- Fix the profile page showing one form's validation errors under all three sections (Profile/Account Security/Danger Zone) at once, plus a related bug where Profile-form errors were silently discarded via redirect.
- Fix books index search erroring when matching by team name (`includes` doesn't force a join for a raw SQL condition).
- Ignore `*.test.js` in ESLint's full-tree scope, matching what `CLAUDE.md` already documents.

## Test plan
- [x] Full Rails suite green (1542 tests, 0 failures) as of the last commit before the search fix; re-verify after merge
- [x] Profile-page fix verified live via Playwright MCP (before/after screenshots in `.playwright-mcp/profile-errors/`)
- [x] Rubocop clean on all touched Ruby files

🤖 Generated with [Claude Code](https://claude.com/claude-code)


app/controllers/concerns/authentication/current_case_manager.rb
```diff
@@ -55,6 +55,16 @@ def check_case
       render json: { message: 'Case not found!' }, status: :not_found unless @case
     end
 
+    # @case is nil when the case doesn't exist, or isn't public/accessible to
+    # this (possibly anonymous) visitor. Render the standard 404 page instead
+    # of letting an HTML-rendering controller/view crash on a nil @case.
+    def render_404_page_unless_case
```
epugh: going to remove this...  new appraoch for centralized 404

PR #1800, opened 2026-09-21
Llm judge provider registry

frutik:
Title: AI Judges: provider registry and adapter seam, with Jev listed as coming soon

  ## Description

  Groundwork for supporting TypeSafe **Jev** as an LLM judge, landed as eight
  behaviour-preserving steps plus two deliberate behaviour changes. No schema
  changes, no new gems.

  * **`LlmProviders` registry** (`app/models/llm_provider.rb`, `llm_providers.rb`) — one
    definition per vendor: label, form defaults, help text, adapter, auth style. The AI
    Judge form's provider dropdown and its `PROVIDER_PRESETS` JavaScript are rendered from
    it (−95 lines of duplicated data in the ERB).
  * **TypeSafe Jev listed as a placeholder** — selecting it shows a "coming soon" banner
    linking to `console.typesafe.ai/keys`, fills and locks the endpoint/model it fixes as
    read-only, and leaves LLM Key editable, so a team can get a key ready. Saving a judge
    pointed at it is refused (controller guard + no adapter registered).
  * **`JudgeScale`** — the book's rating scale as one object (`values`, `label_for`,
    `describe`, `includes?`), shared by the system-prompt augmentation and the job's
    rating check.
  * **`LlmConnection`** — the Faraday/JSON/retry setup, shared by every LLM caller.
  * **`LlmJudgeAdapters::{Base,OpenAi,Anthropic}`** — one class per API dialect, exposing
    `request_envelope(...) -> {path:, headers:, body:}` and
    `apply_response(judgement, body_hash, book:)`. `LlmService` keeps its public surface
    and shrinks to transport: resolve adapter, post, hand the body back.
  * **`JudgementFinalizer`** — "is this rating usable?" in one place (blank → unrateable;
    off-scale → annotate with the raw value, then unrateable), used by the judging job and
    the prompt preview.

  Two intentional behaviour changes, each in its own commit:

  1. **A rate-limited LLM call is now actually retried.** `faraday-retry` only retries
     idempotent methods by default, and every judge call is a POST — so `retry_statuses`
     had never retried anything. `retry_if` now narrows by exception instead of verb:
     a POST refused with 429/529 is retried (the provider produced no completion, so
     nothing is billed or judged twice), while a timeout or connection failure is not,
     because that request may already have been processed.
  2. **The prompt preview applies the same rules as a real run.** Previously you could
     tune a prompt, watch it answer `3` on a 0/1 book, see "LLM Response: 3.0" and think it
     worked — then start a run where every such judgement came back unrateable. The preview
     now shows an **Unrateable** badge and the annotation naming the rejected value.

  Also fixes `User#judge_options=` writing under a symbol key while the reader digs for a
  string one, which produced JSON objects with two `judge_options` keys (last-one-wins
  today; `json` 3.0 raises outright).

  Design decisions and the deployable step order are recorded in
  `docs/adr/0001-llm-judge-provider-architecture.md`; the Jev and (deferred) OpenAI Batch
  implementation plans live in `docs/todo/`.

  ## Motivation and Context

  Adding any provider meant editing four places that had to agree: two constants and two
  `case` statements in `LlmService`, the `<select>` in the AI Judge form, and a
  hand-maintained `PROVIDER_PRESETS` JavaScript object in that same template.

  Jev makes that worse rather than better: it is a typed "System One" model, not a chat
  model — no system prompt, no prose answer, a score over criteria with a probability
  distribution — so it needs its own adapter, and the book's scale becomes part of the
  *request* rather than prose in a prompt. The deferred OpenAI Batch work needs the same
  seam from the other direction: build a request now, read its answer hours later in
  another process.

  Both start by pulling `LlmService` apart along the same line, so it is done once, here,
  while there is a second dialect to prove it against. Jev itself is not implemented in
  this PR — it is listed so teams can see what it will need.

  ## How Has This Been Tested?

  * `bin/docker r rails test test/services test/jobs test/controllers` — 1080 tests,
    3862 assertions, 0 failures. `bin/docker r bundle exec rubocop` clean on every touched
    file.
  * **`test/services/llm_service_test.rb` passes unedited** through the whole refactor —
    400+ lines pinning paths, auth headers and bodies for OpenAI, three Azure variants,
    Anthropic, Gemini and Ollama. Same for `test/jobs/run_judge_judy_job_test.rb`. If those
    had needed editing, the refactor would not have been behaviour-preserving.
  * New suites: `llm_providers_test.rb` (pins the exact per-provider defaults the ERB used
    to hardcode, and that the rendered `PROVIDER_PRESETS` JSON equals the registry),
    `judge_scale_test.rb`, `judgement_finalizer_test.rb`, `llm_connection_test.rb`
    (429 retried then succeeding, 529 retried to exhaustion, 500 and timeouts not retried),
    and `llm_judge_adapters_test.rb`, which tests the batch-readiness contract explicitly:
    building a request does no I/O, the envelope survives a JSON round trip, a response
    applies from a plain Hash, applying never saves — plus a simulation that serializes a
    request as a batch JSONL line, reads it back and applies a canned output body, with no
    batch code in the tree.
  * Manual testing against the local dev stack and Ollama (`qwen3:0.6b`), with
    `docs/manual-testing/tracking.yml` updated: **12.1** (provider dropdown and presets
    unchanged for all 8 existing providers, before/after screenshots identical), **12.2**
    (editing an existing judge still saves, `judge_options` intact), **12.4** (Refine Prompt
    returns a rating; a forced out-of-scale answer renders Unrateable with the annotation),
    **12.5** (bounded Judge Judy runs; both the in-scale and out-of-scale paths hit for real),
    and new scenario **12.6** for the coming-soon provider.

  `.playwright-mcp/` (gitignored) holds before/after captures of the AI Judge form for
  OpenAI and Ollama — identical across the registry change — and the Jev "coming soon"
  state. *Note: a cleanly framed capture of the new Unrateable preview state was not
  obtained; the browser window rendered clipped during the session, so that change was
  verified from the rendered DOM plus controller tests. Happy to add the screenshot.*

  ## Types of changes

  - [ ] New feature (non-breaking change which adds new functionality)

  ## Checklist:


epugh:
Looking at Jev now.

epugh:
@andrew-boyd says "Wherever you’re using LLMs to do classification of known next things, Jev is the answer"

frutik:
@epugh This isn’t Jev yet. This refactoring preserves existing behaviour apart from a few bug fixes and prepares Quepid for Jev (and other not-prompt-based judges). The Jev integration is ready too (in my fork), but it needs this change first.

frutik:
techniclaly https://github.com/o19s/quepid/pull/1802 deprecates that mr @epugh

